PR #25, opened 2021-09-24
Fix u64/UBigInt 'FromSql' trait impl

aarashy:
On line 157 of this file, we are invoking the macro:
```
from_sql_integral!(u64);
```
But the macro doesn't have an arm for ValueRef::UBigInt. 

The consequence is that, while inserting u64 into columns of type UBigInt succeeds, reading them back fails. 

I tested this simple fix locally within my project.


wangfenjin:
Thanks for your contribution!

PR #27, opened 2021-09-27
HugeInt write/readback failures

aarashy:
I noticed that, while writing hugeints, the value I read back from DB was not the same as the value I had written. This is true for all i128s which are larger than i64::MAX. 

I noticed that the Apache Arrow library being used under the hood does not have a DataType::Int128.
The function `value_ref` on `src/row.rs:342` does not have a match arm for handling i128 correctly. Somehow, the upper 64 bits of the HugeInt are definitely being thrown away by duckdb-rs conversions.

One option that should work well would be to use the `DataType::FixedSizeBinary` so that, from Arrow's perspective, you are passing around a 128-bit blob.

@wangfenjin , would you be able to help with making this test succeed? 😸

### DIFF
--- a/src/types/from_sql.rs
+++ b/src/types/from_sql.rs
@@ -313,6 +313,19 @@ mod test {
         Ok(())
     }
 
+    // This test asserts that i128s above/below the i64 max/min can written and retrieved properly.
+    #[test]
+    fn test_hugeint_max_min() -> Result<()> {
+        let db = Connection::open_in_memory()?;
+        db.execute("CREATE TABLE signed_int (u1 hugeint, u2 hugeint);", [])?;
+        let i128max: i128 = i128::MAX;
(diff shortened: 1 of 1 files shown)


wangfenjin:
There are 3 issues here:
1. for the insert, currently we don't support i128 https://github.com/wangfenjin/duckdb-rs/blob/main/src/statement.rs#L424
2. we use rust-decimal to parse the hugeint, it also doesn't support i128::Max https://github.com/paupino/rust-decimal/blob/master/src/constants.rs#L28
3. the min value for hugeint is -170141183460469231731687303715884105727 as state in https://duckdb.org/docs/sql/data_types/numeric

If we want to make this test pass, we need to fix this three issues

wangfenjin:
Create an issue to rust-decimal https://github.com/paupino/rust-decimal/issues/427

PR #30, opened 2022-01-02
Update duckdb to v0.3.1

tachiniererin:
Updated duckdb to the latest upstream release and also dependencies (0.3.1 is compatible with arrow v6 so i bumped this too).
Also fixed the clippy warnings in a second commit.


wangfenjin:
Thank you so much for your contribution!

Issue #6, opened 2021-07-13
Add support for connection pool


wangfenjin:
Maybe use https://github.com/sfackler/r2d2

jaredzhou:
Hi, i want to add a r2d2 connection pool to duckdb-rs, the problem is that when creating multiple connection, the Config need to be cloned multiple times, but duckdb_config can only be dropped once. I think we can use Arc<Option<ffi::duckdb_config>> for config instead, when the Arc strong_count is 1, we can call duckdb_destroy_config method. The problem is we need to change the Connection code with config as well, do you think it is a good idea? or do you have any suggestion @wangfenjin

wangfenjin:
When we clone Config, can we also clone the inner duckdb_config? That way every connection's config is independent.

Thanks for your interests, I'll assign this issue to you. Happy to review the PR!

jaredzhou:
yeah, clone duckdb_config can do the job, i will try it

pricesmith:
Can someone show me how this is done?

PR #32, opened 2022-01-26
add r2d2 connection pool

jaredzhou:
it seems like we dont need to clone the Config, we just need to clone the connection for more connections.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ modern-full = [
     "chrono",
     "serde_json",
     "url",
+    "r2d2",
 ]
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


wangfenjin:
You can try to run `cargo clippy --all-targets --workspace --features bundled --features modern-full -- -D warnings -A clippy::redundant-closure` locally to test

wangfenjin:
Also let's make r2d2 as a feature so that anyone doesn't need connection pool don't need to depends on this.

You can refer to what I did for `chrono`, should be several lines of config change

jaredzhou:
> Also let's make r2d2 as a feature so that anyone doesn't need connection pool don't need to depends on this.
> 
> You can refer to what I did for `chrono`, should be several lines of config change

did not see this , i will fix this

wangfenjin:
Thanks for your great work!

Close #6

jaredzhou:
thank u for the patience

PR #47, opened 2022-04-18
Question about usage of RefCell and Connection being `!Sync`

aarashy:
Hello!

You can ignore the actual diff of this PR for the moment, my real purpose is to raise a discussion.

1. I'm interested in wrapping `Connection` with an `Arc<RwLock<_>>` rather than an `Arc<Mutex<_>>`, so that I may have multiple reader threads simultaneously reading a database. (According to my reading of [1], [2], it's even alright for different **processes** to simultaneously read the same file, but I just want multiple async threads in the same process using the same connection. It doesn't seem like there would be any problem with this from DuckDB's perspective - Though I raised a thread on their discord channel to confirm this).

2. I noticed that `conn.execute_batch` takes a `&self` - immutable ref instead of mutable ref - which seemed odd to me, since it does mutate the underlying data. I realized this is possible due to the usage of RefCell and `borrow_mut` around `InnerConnection` which this crate took from rusqlite which was our starting point. This led me to a rusqlite thread where they discuss intentionally opting out of the compile-time borrow checker [3]. 

I realize that there are limitations of using the compile-time borrow checker... there's clearly not a practical/efficient way to check whether a raw SQL string is read-only or not... but the usage of RefCell makes our `Connection` struct `!Sync`. This means that I cannot pass `Arc<RwLock<Connection>>` into a `tokio::spawn_blocking` because only Mutex is the only wrapper that can make a non-`Sync` T into `Sync`.

So basically, my question is: do you think it's possible and/or desirable to make `Connection` into a `Sync` struct by avoiding usage of RefCell?
If I were to embark on that, do you think I should take special care to continue supporting the sqlite API's reference mutability?

The long term goal would be to make `query_arrow` take a immutable ref instead of a mut ref so that multiple Tokio threads can process SELECT statements on the same connection.

[1]: https://github.com/duckdb/duckdb/issues/1343
[2]: https://github.com/duckdb/duckdb/issues/77
[3]: https://github.com/rusqlite/rusqlite/pull/693

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ struct Person {
 }
 
 fn main() -> Result<()> {
-    let conn = Connection::open_in_memory()?;
+    let mut conn = Connection::open_in_memory()?;
 
     conn.execute_batch(
         r"CREATE SEQUENCE seq;
--- a/src/appender.rs
+++ b/src/appender.rs
@@ -22,7 +22,7 @@ impl Appender<'_> {
     ///
     /// ```rust,no_run
     /// # use duckdb::{Connection, Result, params};
-    /// fn insert_rows(conn: &Connection) -> Result<()> {
+    /// fn insert_rows(conn: &mut Connection) -> Result<()> {
     ///     let mut app = conn.appender("foo")?;
     ///     app.append_rows([[1, 2], [3, 4], [5, 6], [7, 8], [9, 10]])?;
     ///     Ok(())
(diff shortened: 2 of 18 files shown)


wangfenjin:
I'm not an expert of rust concurrency ecosystem, but from what I know, you may misunderstand about concurrent connection. They don't mean read concurrently in one connection, but mean multiple connections to the same db instance. We can't read concurrently in one connection, think about multiple reader from 1 TCP connection, that's just impossible.

Currently we support Connection.clone(), and also implement [connection pool](https://github.com/wangfenjin/duckdb-rs/blob/main/src/r2d2.rs), should support your needs. All you need is to clone the connection and move into the tokio process. It should work.

But after read the link you provided, we do need to make the cloned connection read-only.

aarashy:
Thanks for the response. It seems like cloning the connection is cheap enough, so maybe that (or just opening new read-only connections) really is the best way to go.

PR #54, opened 2022-05-18
Remove panic from inner connection initialization

aarashy:
It seems the main reason a `panic!` remains in `InnerConnection::new` was because the function signature is copied from rusqlite. In their case, this function was actually infallible. 

It's dangerous to leave this `panic!` in the code here - I have a hunch it was responsible for my host process crashing in https://github.com/duckdb/duckdb/issues/3666, probably due to attempting to connect to a corrupted database file.

If we have to choose between keeping their function signature or making this code safer, we ought to choose to make our code safer. 

The unfortunate side affect is that we can't impl `Clone` anymore for connection because creating a new connection is now fallible (Or rather, it already was fallible, but the use of `panic!` hid this from the type checker).

### DIFF
--- a/src/inner_connection.rs
+++ b/src/inner_connection.rs
@@ -16,25 +16,19 @@ pub struct InnerConnection {
     owned: bool,
 }
 
-impl Clone for InnerConnection {
-    fn clone(&self) -> Self {
-        unsafe { InnerConnection::new(self.db, false) }
-    }
-}
-
 impl InnerConnection {
-    #[allow(clippy::mutex_atomic)]
     #[inline]
-    pub unsafe fn new(db: ffi::duckdb_database, owned: bool) -> InnerConnection {
+    pub unsafe fn new(db: ffi::duckdb_database, owned: bool) -> Result<InnerConnection> {
         let mut con: ffi::duckdb_connection = ptr::null_mut();
         let r = ffi::duckdb_connect(db, &mut con);
         if r != ffi::DuckDBSuccess {
             ffi::duckdb_disconnect(&mut con);
-            let e = Error::DuckDBFailure(ffi::Error::new(r), Some("connect error".to_owned()));
-            // TODO: fix this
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -924,7 +923,7 @@ mod test {
     fn test_clone() -> Result<()> {
         let owned_con = checked_memory_handle();
         {
-            let cloned_con = owned_con.clone();
+            let cloned_con = owned_con.try_clone().unwrap();
             cloned_con.execute_batch("PRAGMA VERSION")?;
         }
         owned_con.close().unwrap();
```
aarashy, line 929: I looked into the implementation of `close` -- Isn't there a bug in the usage of `owned` and `close()` here?

The initial object has `owned = true` and the cloned one has `owned = false`. If we drop the original before dropping the clone (opposite order of this test) wouldn't the original actually close the DuckDB file and leave the clone with a dangling reference?

It seems like we need a shared reference counter to close the DB correctly, rather than this `owned` boolean.

wangfenjin:
I'm OK with this change. Could you fix the CI issue?

src/lib.rs
wangfenjin: It could be a problem, could you write a test about it?

src/inner_connection.rs
```diff
@@ -71,6 +64,11 @@ impl InnerConnection {
         Ok(())
     }
 
+    /// Creates a new connection to the already-opened database.
+    pub fn try_clone(&self) -> Result<Self> {
+        unsafe { InnerConnection::new(self.db, false) }
+    }
```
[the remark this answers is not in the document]
wangfenjin: the API is fine to me, what we may need to do is set the connection to readonly. but seems there is no support of this in the c api.

But you can add test to verify if the cloned connection can write or not, as the default AccessMode is AUTOMATIC, maybe they already handled that.

PR #63, opened 2022-06-30
Compile from source on Windows

### LINKED ISSUE #62: Unable to compile `duckdb-sys` on Windows
We were setting up Windows builds in GitHub Actions for our app and noticed that the `duckdb-sys v0.4.0` crate's build script is unable to compile DuckDB from source.

There are a lot of suspicious warnings, and I've had to attach the output as a file because GitHub has a  65536 character limit on comments, but I think this is the thing that actually fails the build:

```
fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
```

[(Link to full output from the build script)](https://github.com/wangfenjin/duckdb-rs/files/9007121/stderr.txt)

I notice you've actually commented out the Windows build for this repo's CI. Does that mean this is a known issue?

https://github.com/wangfenjin/duckdb-rs/blob/30d8e3e637f8a3f9b96ca6de9e2275bc69a5bbae/.github/workflows/rust.yaml#L15-L17

Michael-F-Bryan:
We found that you can compile the `duckdb-rs` crate with its `bundled` feature on Windows if you use Clang as the compiler (i.e. set `CC=clang` and `CXX=clang++`). All we needed to do is let the `cc` crate decide whether to emit `-fPIC` instead of forcing it to always be enabled (the default is `false` for Windows, `true` for everything else).

Fixes #62.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          #- { target: x86_64-pc-windows-msvc, os: windows-latest }
+          - { target: x86_64-pc-windows-msvc, os: windows-latest }
           - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
           #- { target: x86_64-apple-darwin, os: macos-latest }
           #- {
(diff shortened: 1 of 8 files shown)


wangfenjin:
@Michael-F-Bryan Thanks for your effect on this! I had create another PR based on yours #65

PR #77, opened 2022-09-07
Bump 0.4.0 -> 0.5.0

chhetripradeep:
DuckDB 0.5.0 was released recently https://github.com/duckdb/duckdb/releases/tag/v0.5.0

Used upgrade.sh script for this PR.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -38,7 +38,7 @@ jobs:
         name: Download duckdb
         with:
           repository: "duckdb/duckdb"
-          tag: "v0.4.0"
+          tag: "v0.5.0"
           fileName: ${{ matrix.duckdb }}
           out-file-path: .
 
--- a/.gitignore
+++ b/.gitignore
@@ -15,8 +15,9 @@ Cargo.lock
 /target
 **/target
 
-# Ignore vscode
+# Ignore vscode and intellij
 .vscode
+.idea
 
 # Ignore zip file
 *.zip
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "duckdb"
-version = "0.4.0"
+version = "0.5.0"
 authors = ["wangfenjin <wangfenj@gmail.com>"]
 edition = "2021"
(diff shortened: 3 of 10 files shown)


wangfenjin:
Hi @chhetripradeep , thank you so much for working on this. The CI might report some error, please help follow up.

You can run `cargo test --features bundled --features modern-full -- --nocapture` locally to find the breaking test.

chhetripradeep:
Hi @wangfenjin Thank you for the quick review. Sure, I will follow up on the error.

wangfenjin:
Acctually when I prepare the upgrade, I found an issue https://github.com/duckdb/duckdb/issues/4612 , although it's fixed in master version, but still the 0.5.0 version will break the test, fixing it you can just delete the `where id = ` clause for now.

So recently I'm hesitate if we still publish the new version... But as you create this PR, I think I'll still publish the new version.

wangfenjin:
Also you noticed that #76 arrow dependency also break because of https://github.com/apache/arrow-rs/issues/2659

So if I release the new version, may stick to 21 for now.

wangfenjin:
For test_ddl can change to these

```rust
    fn test_ddl() -> Result<()> {                                                                                                                                                                [27/2402]
        let db = Connection::open_in_memory()?;
        db.execute_batch(
            r#"
            CREATE TABLE foo (x INT);
            INSERT INTO foo VALUES (1);
        "#,
        )?;

        let sql = "SELECT * FROM foo";

        {
            let mut stmt = db.prepare_cached(sql)?;
            assert_eq!(Ok(Some(1i32)), stmt.query([])?.map(|r| r.get(0)).next());
        }

        db.execute_batch(
            r#"
            ALTER TABLE foo ADD COLUMN y INT;
            UPDATE foo SET y = 2;
        "#,
        )?;

        {
            // Rebinding statement after catalog change resulted in change of types
            let mut stmt = db.prepare_cached(sql)?;
            assert_eq!(
                Ok(Some((1i32, 2i32))),
                stmt.query([])?.map(|r| <(i32, i32)>::try_from(r)).next()
            );
        }
        Ok(())
    }
```

chhetripradeep:
Thank you for the help

PR #93, opened 2022-10-24
Re-exports arrow dependency to minimise version maintenance for crate…

### LINKED ISSUE #92: Re-export dependencies (in particular arrow) to avoid version incompatibilties in client crates
Hi,

The Rust Arrow crate is used in many projects through the Rust data ecosystem (duckdb-rs, DataFusion, Polars, connectorx, ...). They also rapidly cycle through major version numbers (currently v25 and I think they were on 19 when I started a few months ago). This makes it extremely difficult if not impossible to find common version numbers to use in any client crate if you want to use more than one of these projects in one crate.

The only way around this that I have found so far is to use a re-exported version from the place where it is used. DataFusion does this for example here: https://github.com/apache/arrow-datafusion/blob/572c20e904b48ea4819210b983d41b1b08c23b46/datafusion/core/src/lib.rs#L232

Please can you add the same top-level re-export to duckdb-rs? It has to be the `arrow` crate at the top-level because I use a number of items from it.

This is probably a one line change and I'm happy to submit a PR for it if you like.

snth:
Hi,

I think this is all that is required to fix #92.

For my use case it would be ideal if you could add `parquet` as well but I understand if you are hesitant to add a dependency to the crate that you're not even using. However `arrow` and `parquet` are often used together so if you want to write the results of your DuckDB query out to a parquet file then you might well run into the same versin dependency problem if you don't have access to the same version of `arrow` and `parquet`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,17 @@ forked from rusqlite as duckdb also tries to expose a sqlite3 compatible API.
 
 ```rust
 use duckdb::{params, Connection, Result};
-use arrow::record_batch::RecordBatch;
-use arrow::util::pretty::print_batches;
+
+// In your project, we need to keep the arrow version same as the version used in duckdb.
+// Refer to https://github.com/wangfenjin/duckdb-rs/issues/92
+// You can either:
+use duckdb::arrow::record_batch::RecordBatch;
+// Or in your Cargo.toml, use * as the version; features can be toggled according to your needs
+// arrow = { version = "*", default-features = false, features = ["prettyprint"] }
+// Then you can:
+// use arrow::record_batch::RecordBatch;
+
+use duckdb::arrow::util::pretty::print_batches;
 
 #[derive(Debug)]
 struct Person {
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -2,8 +2,8 @@
 
 extern crate duckdb;
 
-use arrow::record_batch::RecordBatch;
-use arrow::util::pretty::print_batches;
+use duckdb::arrow::record_batch::RecordBatch;
+use duckdb::arrow::util::pretty::print_batches;
 use duckdb::{params, Connection, Result};
 
 #[derive(Debug)]
--- a/examples/parquet.rs
+++ b/examples/parquet.rs
@@ -1,8 +1,7 @@
 extern crate duckdb;
+use duckdb::arrow::record_batch::RecordBatch;
+use duckdb::arrow::util::pretty::print_batches;
 use duckdb::{Connection, Result};
-
-use arrow::record_batch::RecordBatch;
-use arrow::util::pretty::print_batches;
 
 fn main() -> Result<()> {
     let db = Connection::open_in_memory()?;
--- a/src/arrow_batch.rs
+++ b/src/arrow_batch.rs
@@ -1,5 +1,5 @@
+use super::arrow::{datatypes::SchemaRef, record_batch::RecordBatch};
 use super::Statement;
-use arrow::{datatypes::SchemaRef, record_batch::RecordBatch};
 
 /// An handle for the resulting RecordBatch of a query.
 #[must_use = "Arrow is lazy and will do nothing unless consumed"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,8 +3,8 @@
 //!
 //! ```rust
 //! use duckdb::{params, Connection, Result};
-//! use arrow::record_batch::RecordBatch;
-//! use arrow::util::pretty::print_batches;
+//! use duckdb::arrow::record_batch::RecordBatch;
+//! use duckdb::arrow::util::pretty::print_batches;
 //!
 //! #[derive(Debug)]
 //! struct Person {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -87,6 +87,9 @@ pub use crate::statement::Statement;
 pub use crate::transaction::{DropBehavior, Savepoint, Transaction, TransactionBehavior};
 pub use crate::types::ToSql;
 
+// re-export dependencies from arrow-rs to minimise version maintenance for crate users
+pub use arrow;
+
 #[macro_use]
 mod error;
 mod appender;


snth:
I've added a second commit with the additional features that I would need. What do you think?

snth:
Actually we can guard this with a feature flag so that it won't affect your users that aren't interested in this. I'll implement that tomorrow.

wangfenjin:
For parquet if we don’t use the same version it will also report error?

wangfenjin:
Currently solution seems weird to me as we even need to export unused features / crete.

The temporary solution I may suggest for now is update the README to tell users to use `*` as the version so cargo can figure out the version by itself.

```rust
# use * as the version; features can be toggled according to your needs
arrow = { version = "*", default-features = false, features = ["prettyprint"] }
```

snth:
You are right, it is quite weird and I totally understand that you probably don't want to burden your users with the weight of additional dependencies that aren't even used in your crate. I think I may have thought of a solution to my problem using Rust workspaces so that the parquet dependency won't be necessary on your side. I still think that reexporting the arrow module would be good practice though. I"ll update the PR to remove the parquet dependency.

---
If I have time I'll try to explain why your suggested `version="*"` solution won't work for me though.

snth:
I've updated the PR with just the `arrow` reexport. I think this is good practice and will be useful for your other users as well.

---
The arrow version problem is as follows:

In my crate I would like to use [duckdb-rs](https://docs.rs/duckdb/latest/duckdb/), [datafusion](https://docs.rs/datafusion/latest/datafusion/) and [connectorx](https://docs.rs/connectorx/latest/connectorx/).

If you look at those docs.rs links, you'll see that they all use [arrow](https://docs.rs/arrow/latest/arrow/) and the versions of arrow used are 23, 24, and 9 respectively. So even if you put `version="*"`, there is no version that satisfies all those simultaneously. I looked through the docs.rs docs for previous versions of each package to see if I could use some old version of each one so that they could all match on the same arrow version but I couldn't find a common version.

Now DataFusion reexports arrow so instead writing `use arrow;` in that module, I could write `use datafusion::arrow;`. That meant that in my Cargo.toml file I could specify the version of arrow that matched the version used in duckdb-rs. Unfortunately there is no space to specify a separate version for `connectorx` so at the moment I can't use `connectorx`. If you add the `pub use arrow;` line from this PR, then I can use `use duckdb::arrow;` in my duckdb module which means it frees up my Cargo.toml version of arrow to be the same as that used in `connectorx` (9) and then I can use connectorx again.

What I want to do with arrow though is to write the data to parquet files and unfortunately the parquet crate is tightly coupled to arrow and uses the same version numbers. That means though that unfortunately I will need to have access to parquet versions 23 and 9 in the same crate if I want to be able to write data from duckdb-rs or connectorx to parquet. Therefore even though you don't actually use parquet in your crate, it still leaves me with a problem.

---
Anyway, I think I may have found a solution around this problem by creating library crates in my main crate using the workspaces feature. Those library crates will have only one purpose and that is to reexport the correct version of arrow and parquet to my main crate. I'll see if that solves the problem.

wangfenjin:
One of the reason arrow cause a lot of issue is they always bump big version in each release…

Have you considering https://github.com/jorgecarleitao/arrow2 ? I always think about migrate the dependency to arrow2 but still thinking about the impact

wangfenjin:
Thanks for the contribution, I'll merge the PR

snth:
> One of the reason arrow cause a lot of issue is they always bump big version in each release…

I completely agree with you. I asked about this on the arrow rust Discord and Slack channels and the reply I got is that this works well for them 🤷‍♂️. I don't really understand that because I think it causes a lot of problems in the Rust Data ecosystem. How many breaking changes are there really between each Arrow release? So far all my code has worked unchanged across all arrow versions since about the mid-teens.

> Have you considering https://github.com/jorgecarleitao/arrow2 ? I always think about migrate the dependency to arrow2 but still thinking about the impact

I started looking into arrow2 in my connectorx branch. However it's a bit confusing because they use different terminology 
(Chunk vs RecordBatch ...) and I'm not really that far yet in translating my arrow code to arrow2. There are less examples of arrow2 (I think) so it's also a bit harder to get going on. I would also rather maintain one version of code that works across all query engines, even if they use different versions of arrow as long as I can work around the version problems.

Of course I don't think DataFusion supports arrow2 or is likely to. It's really unfortunate that there is this split developing ...

wangfenjin:
@snth actually they do have an arrow2 branch, but the master branch still based on arrow

https://github.com/apache/arrow-datafusion/tree/arrow2

snth:
Oh interesting. There is quite a lot overlap between the arrow and datafusion teams, both in terms of people as well as versioning methodology, so I wasn't expecting this. Thanks @wangfenjin!

wangfenjin:
Haha if you search their issue list they have heavy discussion about the future. One thing they all agree is arrow2 is better 😂 but because licensing issue and some other issue they are still maintaining arrow-rs

mike-luabase:
Now that the crate is re-exported, how do you handle the "mismatched types"?

```
error[E0308]: mismatched types
[0]    --> src/duckdb.rs:187:53
[0]     |
[0] 187 |             let arr_col_str = array_value_to_string(col, 1);
[0]     |                               --------------------- ^^^ expected trait `arrow::array::Array`, found trait `duckdb::arrow::array::Array`
[0]     |                               |
[0]     |                               arguments to this function are incorrect
[0]     |
[0]     = note: expected reference `&Arc<(dyn arrow::array::Array + 'static)>`
[0]                found reference `&Arc<dyn duckdb::arrow::array::Array>`
[0]     = note: perhaps two different versions of crate `arrow_array` are being used?
[0] note: function defined here
[0]    --> /Users/mritchie712/.cargo/registry/src/github.com-1ecc6299db9ec823/arrow-cast-27.0.0/src/display.rs:329:8
[0]     |
[0] 329 | pub fn array_value_to_string(
[0]     |        ^^^^^^^^^^^^^^^^^^^^^
[0] 
error[E0277]: the trait bound `duckdb::arrow::array::ArrayData: arrow::array::Array` is not satisfied
[0]    --> src/duckdb.rs:186:29
[0]     |
[0] 186 |             let got_arrow = col.data() as &dyn arrow::array::Array;
[0]     |                             ^^^^^^^^^^ the trait `arrow::array::Array` is not implemented for `duckdb::arrow::array::ArrayData`
[0]     |
[0]     = help: the following other types implement trait `arrow::array::Array`:
[0]               &'a T
[0]               Arc<(dyn arrow::array::Array + 'static)>
[0]               arrow::array::BooleanArray
[0]               arrow::array::DictionaryArray<T>
[0]               arrow::array::FixedSizeBinaryArray
[0]               arrow::array::FixedSizeListArray
[0]               arrow::array::GenericByteArray<T>
[0]               arrow::array::GenericListArray<OffsetSize>
[0]             and 6 others
[0]     = note: required for the cast from `duckdb::arrow::array::ArrayData` to the object type `dyn arrow::array::Array`
```

mike-luabase:
nm, I see, you need to use:

```
use duckdb::arrow::util::display::array_value_to_string;
```

Issue #87, opened 2022-10-09
Problem with `stmt.query_arrow([])?.collect()`


sspaeti:
I'm a total starter and trying to learn Rust with DuckDB. Just close if it's too basic of a question, just posting it here in case others have trouble getting started.

Error: 
```
rustc: a value of type `Vec<RecordBatch>` cannot be built from an iterator over elements of type `arrow::record_batch::RecordBatch` the trait `FromIterator<arrow::record_batch::RecordBatch>` is not implemented for `Vec<RecordBatch>`
```
Question: What am I doing wrong:
![image](https://user-images.githubusercontent.com/689199/194746499-ebb08e69-7cb5-48e0-8333-7b3e8ffb5299.png)

I understand the rust `iterator` does not like the arrow `RecordBatch`. But how can I fix that, did I use the wrong arrow dependencies in my `Cargo.toml` on the right?

wangfenjin:
why depends on duckdb 0.1 version but libduckdb-sys 0.5.1 ? suggest to always use the same version, and latest version which is 0.5.1.

0.1 version duckdb doesn't support arrow

sspaeti:
Thank you so much for your kind answer. OK now I see, I updated to the latest `0.5.1`, but still has the same error. 

Which `Cargo.toml` should be used for the simple example on the left? Should that work with the right dependencies, or should I follow something like this https://github.com/wangfenjin/duckdb-rs/blob/main/Cargo.toml?

wangfenjin:
Refer to #88 , the readme example should works fine. You can try run the example in command line with cargo. Not sure what's the problem with your code editor

sspaeti:
Didn't really manage to make it work, but I used some part of this https://github.com/hotg-ai/weld-studio/blob/7121cbffbbe7dd4ee703b1dd8b3be89daee1f857/src-tauri/src/sql.rs and removed `let rbs: Vec<RecordBatch> = stmt.query_arrow([])?.collect();`.

Now I have managed to have a working starting point. Thanks so much; I will close.

wangfenjin:
Hi @sspaeti 

I checked the issue again. The arrow version in your project's Cargo.toml should be same as the version in duckdb (latest one is version 23) . Otherwise it may report error.

Sorry for the inconvenience.

snth:
[...]

PR #94, opened 2022-10-25
Add comments on arrow versoin

### LINKED ISSUE #92: Re-export dependencies (in particular arrow) to avoid version incompatibilties in client crates

wangfenjin:
Fix #92

Related to #87 #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ forked from rusqlite as duckdb also tries to expose a sqlite3 compatible API.
 
 ```rust
 use duckdb::{params, Connection, Result};
+// In your Cargo.toml, use * as the version; features can be toggled according to your needs
+// https://github.com/wangfenjin/duckdb-rs/issues/92
+// arrow = { version = "*", default-features = false, features = ["prettyprint"] }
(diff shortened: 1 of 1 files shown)


wangfenjin:
Prefer #93

PR #105, opened 2022-12-10
Add timestamp wrapper type

0b01:
This PR adds a Timestamp wrapper type for appending timestamps in the appender. This should be the lowest overhead way for inserting timestamps.

### DIFF
--- a/src/appender.rs
+++ b/src/appender.rs
@@ -235,13 +235,14 @@ mod test {
         let db = Connection::open_in_memory()?;
         db.execute_batch("CREATE TABLE foo(x TIMESTAMP)")?;
 
+        let d = Duration::from_secs(1);
         {
             let mut app = db.appender("foo")?;
-            app.append_row([Duration::from_secs(1)])?;
+            app.append_row([d])?;
         }
 
-        let val = db.query_row("SELECT x FROM foo", [], |row| <(i32,)>::try_from(row))?;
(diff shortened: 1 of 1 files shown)


src/types/mod.rs
```diff
@@ -85,6 +85,30 @@ mod url;
...
+    Second(i64),
+    /// Millisecond
+    Millisecond(i64),
+    /// Microsecond
+    Microsecond(i64),
+    /// Nanosecond
+    Nanosecond(i64),
+}
+
+impl ToSql for Timestamp {
```
wangfenjin, line 101: Seems it's more reasonable to `impl ToSql for std::time::Duration` or `impl ToSql for std::time::Instant` ?

https://doc.rust-lang.org/std/time/struct.Instant.html
https://doc.rust-lang.org/std/time/struct.Duration.html

wangfenjin: And we'd better move the logic into https://github.com/wangfenjin/duckdb-rs/blob/main/src/types/to_sql.rs#L204

wangfenjin (commented):
Thanks for the contribution! There is one comment bellow

src/types/mod.rs
wangfenjin: The reason is if we define our own Timestamp type, users need to convert their type to ours

PR #110, opened 2022-12-14
Support of unsigned types

### LINKED ISSUE #109: Incorrect binding to unsigned integers
Even though DuckDB supports unsigned integer, this crate still binds values as signed integers, causing overflow error:

```rust
use duckdb::*;
fn main() {
    let conn = Connection::open_in_memory().expect("open db");
    conn.execute_batch(
        r"CREATE TABLE a (x UBIGINT);").expect("create table");
    conn.execute("INSERT INTO a(x) VALUES (?)", params![0xFFFFFFFFFFFFFFFFu64]).expect("insert");
}
```

Output:

```
thread 'main' panicked at 'insert: DuckDBFailure(Error { code: Unknown, extended_code: 1 }, Some("Conversion Error: Type INT64 with value -1 can't be cast because the value is out of range for the destination type UINT64"))', src/main.rs:7:81
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This behavior might be inherited from Rusqlite, because I know Rusqlite does something like this.

rongcuid:
Fixes #109

### DIFF
--- a/src/statement.rs
+++ b/src/statement.rs
@@ -648,9 +648,13 @@ mod test {
             db.execute_batch("CREATE TABLE test (name TEXT, value UINTEGER)")?;
             let mut stmt = db.prepare("INSERT INTO test(name, value) VALUES (?, ?)")?;
             stmt.raw_bind_parameter(1, "negative")?;
-            stmt.raw_bind_parameter(2, 0xFFFFFFFFu32)?;
+            stmt.raw_bind_parameter(2, u32::MAX)?;
             let n = stmt.raw_execute()?;
             assert_eq!(n, 1);
+            assert_eq!(
+                u32::MAX,
+                db.query_row::<u32, _, _>("SELECT value FROM test", [], |r| r.get(0))?
+            );
         }
 
         {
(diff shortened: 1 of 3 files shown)


rongcuid:
Also fixed appender

src/statement.rs
```diff
@@ -639,6 +643,26 @@ mod test {
             assert!(rows.next()?.is_none());
         }
 
+        {
+            let db = Connection::open_in_memory()?;
+            db.execute_batch("CREATE TABLE test (name TEXT, value UINTEGER)")?;
+            let mut stmt = db.prepare("INSERT INTO test(name, value) VALUES (?, ?)")?;
+            stmt.raw_bind_parameter(1, "negative")?;
+            stmt.raw_bind_parameter(2, 0xFFFFFFFFu32)?;
```
wangfenjin, line 651: https://doc.rust-lang.org/std/primitive.u32.html#associatedconstant.MAX

src/statement.rs
```diff
@@ -639,6 +643,26 @@ mod test {
...
+        }
+
+        {
+            let db = Connection::open_in_memory()?;
+            db.execute_batch("CREATE TABLE test (name TEXT, value UBIGINT)")?;
+            let mut stmt = db.prepare("INSERT INTO test(name, value) VALUES (?, ?)")?;
+            stmt.raw_bind_parameter(1, "negative")?;
+            stmt.raw_bind_parameter(2, 0xFFFFFFFFFFFFFFFFu64)?;
+            let n = stmt.raw_execute()?;
+            assert_eq!(n, 1);
```
wangfenjin, line 663: Same as u32, please add assert on the value

wangfenjin (commented):
Thanks for the contribution. Few minor comments:

Issue #102, opened 2022-11-28
How to register table function?


Berrysoft:
I've seen from the C API that duckdb could register table function. Would you consider providing such feature?

wangfenjin:
Yes, still haven’t got time to do it

Issue #79, opened 2022-09-09
Example of using Arrow C interface


kesavkolla:
Can someone provide an example of how to seemlessly pass the arrow to duckdb in rust? I have the location of parquet file(s) in Azure ADLSv2. I am currently using object_store and arrow-rs to read those parquet files. I do see duckdb and arrow work very seemless in Python. Is there some example that someone can point how to use in rust.

Thanks in advance.

wangfenjin:
you mean something like this? https://duckdb.org/docs/guides/python/sql_on_arrow

Better share your python examples

kesavkolla:
Yes that is correct. The example given in that URL. Ideally the recordbatchreader querying on it would be super useful from rust.

wangfenjin:
Haven't support that in rust api. For now what you can do is load parquet directly into duckdb

#80

kesavkolla:
Hmmm... unfortunately I can't load parquet in duckdb because duckdb doesn't have support for Azure ADLS. Currently duckdb has no pluggable mechanism for loading parquet files.

Is it possible to write an extension for duckdb in rust?

wangfenjin:
What about download parquet first and then load it into duckdb?

I haven't spent too much time to dive deep into the duckdb extension, need time to investigate

kesavkolla:
Yeah I could do that. Ideally it would be nice to provide Arrow C interface so we can easily integrate between rust and duckdb

wangfenjin:
Yes, the API would be much simpler. 

But the logic still the same, download the file, and load into duckdb. Read into arrow and then bind the data into duckdb might be slower than directly read parquet file into duckdb

kesavkolla:
I am looking at this example in Python:


```python
from adlfs import AzureBlobFileSystem
import duckdb
import pyarrow.parquet as pq
import pyarrow.dataset as ds

account_name = "xxxxx"
account_key = "xxxxx"
abfs = AzureBlobFileSystem( account_name = account_name, account_key = account_key, container_name = "data")
pqdata = ds.dataset("path/inside/abfs", filesystem=abfs)


conn = duckdb.connect(":memory:")
conn.execute("SELECT * from pqdata LIMIT 10")
```

Something similar in rust would be greatly helpful

wangfenjin:
[...]

PR #112, opened 2022-12-19
[WIP] Initial support of table function

wangfenjin:
This is based on @Mause 's https://github.com/Mause/duckdb-extension-framework

It works (refer test_table_function), but I'm thinking about design a more safe and high level API, similiar to [rusqlite vtab](https://github.com/rusqlite/rusqlite/blob/master/src/vtab/mod.rs), So need more time to merge the MR

If anyone interested to work on this base on this version can let me know

related to #102 #79

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,8 @@ arrow = { version = "29", default-features = false, features = ["prettyprint", "
 rust_decimal = "1.14"
 strum = { version = "0.24", features = ["derive"] }
 r2d2 = { version = "0.8.9", optional = true }
+num-derive = { version = "0.3.3" }
+num-traits = { version = "0.2.15" }
 
 [dev-dependencies]
 doc-comment = "0.3"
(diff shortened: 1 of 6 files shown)


wangfenjin:
CI failed because of documentation

Some notes about next steps:

1. [value](https://duckdb.org/docs/api/c/api#value-interface) and related API are don't directly related to table function, but will be used if user want to implement a table function, because user need these API to generate the result set. So we need to make these class of API easy to use
2. [table-functions](https://duckdb.org/docs/api/c/api#table-functions) and related API are used for register the table function, we should design the API simliar to [Module](https://github.com/rusqlite/rusqlite/blob/master/src/vtab/mod.rs#L81) and [VTab](https://github.com/rusqlite/rusqlite/blob/master/src/vtab/mod.rs#L246), to make it easy for user the define the table function as normal rust code, with less unsafe code.
3. We can implement an arrow_scan() table function as an show case and also to verify our API design is good enough

changhiskhan:
@wangfenjin we're working on a rust implementation of lance data format and would be interested in seeing if your PR here can provide an easy way for us to write a Lance Scanner (comes out as Arrow) + replacement scan function (so basically item 3 in your next steps). Are you planning to substantially change the API here?

wangfenjin:
in favor of #138

PR #127, opened 2023-03-07
Rework bundled compilation to support included extensions

ankrgyl:
This change reworks the bundled compilation process so that we can support built-in extensions, like httpfs, to the compilation process. There are a number of changes that support it:

* Reuse `package_build.build_package` from the DuckDB repo to gather source files, including extensions. We no longer use the amalgamation, and end up pulling more than 3,000 files. To keep the number of files in the repo small, we tar.gz the duckdb directory (and untar it on demand). This is all managed from a new script called `update_sources.py`.
* DuckDB is now a submodule of the repo, so that when we want to release a new version, we simply update the repo's submodule and run the script.
* Copy the openssl-finding code from [rust-openssl](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/find_normal.rs), and invoke it when including the httpfs feature
* Add a set of new features (1 per duckdb extension) that control what gets bundled in. We can extend this to include all of the builtin plugins to duckdb.


ankrgyl:
CC @wangfenjin

wangfenjin:
You can run `sh add_rust_fmt_hook.sh` in the folder to make sure code will be formatted before commit

https://github.com/wangfenjin/duckdb-rs/blob/main/add_rustfmt_hook.sh

wangfenjin:
Do you think it's better to depends on the openssl crate directly? Instead of copy all the code to our repo

ankrgyl:
> You can run sh add_rust_fmt_hook.sh in the folder to make sure code will be formatted before commit

Awesome, I've done that, but for some reason I think it's still producing slightly different results (like maybe the line lengths are different). Let's see if this most recent CI passes...

> Do you think it's better to depends on the openssl crate directly? Instead of copy all the code to our repo

I don't think that's possible unfortunately — I had to make some modifications including returning some values out of the `main` function. We could potentially upstream some changes to that crate that make this logic importable, but I'd prefer to separate that from this effort.

wangfenjin:
Let's add a feature maybe called `extension-full` to include all the extensions:
https://github.com/wangfenjin/duckdb-rs/blob/main/Cargo.toml#L27

And then add the flag in ci to make it build passed
https://github.com/wangfenjin/duckdb-rs/blob/main/.github/workflows/rust.yaml#L108-L131

Thanks @ankrgyl

wangfenjin:
It will be better if you can also help add some tests to verify these extensions actually works, the tests will also be used as examples for other users.

The tests can be enabled by `#[cfg(feature = "foo")]`

wangfenjin:
If you don't have the time to add tests it's fine, I'll do that later after this one merged

wangfenjin:
Is the submodule and tar.gz v0.7.1 ?

wangfenjin:
https://github.com/wangfenjin/duckdb-rs/blob/main/libduckdb-sys/upgrade.sh#L14-L16 we may also update this file to make update version easier in the future. But that fine I can do that after this one get merged

snth:
Thank you @ankrgyl . This looks great and I'll be your first happy user of this as I've been hoping for this functionality for some time.

I especially like what you did with tar.gz on the duckdb directory, nice!

ankrgyl:
> If you don't have the time to add tests it's fine, I'll do that later after this one merged

Would prefer if you could do that in a separate change. Thank you for offering!

> Is the submodule and tar.gz v0.7.1 ?

Yes although I noticed that yours are generated on an x86 linux machine and mine on an arm64 mac. Maybe we should generate them on one of your machines?

> https://github.com/wangfenjin/duckdb-rs/blob/main/libduckdb-sys/upgrade.sh#L14-L16 we may also update this file to make update version easier in the future. But that fine I can do that after this one get merged

> https://github.com/wangfenjin/duckdb-rs/blob/main/libduckdb-sys/upgrade.sh#L14-L16 we may also update this file to make update version easier in the future. But that fine I can do that after this one get merged

I think the new `update_sources.py` script should encompass all of the functionality from this script. Could you verify that?

wangfenjin:
> Yes although I noticed that yours are generated on an x86 linux machine and mine on an arm64 mac. Maybe we should generate them on one of your machines?

I generated on a intel mac. Never notice the difference of this, maybe we should delete the file and always generate the file on the fly in the build process. I'll follow up this issue

ankrgyl:
>  I generated on a intel mac. Never notice the difference of this, maybe we should delete the file and always generate the file on the fly in the build process. I'll follow up this issue

Ah got it. I think technically it may be ok (?) at least according to CI, but I agree that generating it dynamically is safer.

ankrgyl:
@wangfenjin could you help me with windows? I'm not familiar nor have an environment of my own to test with.

I followed https://github.com/sfackler/rust-openssl/blob/master/.github/workflows/ci.yml#L93 to add VCPKG and hoped that would work...

.github/workflows/rust.yaml
```diff
@@ -101,11 +101,19 @@ jobs:
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v2
+      - run: echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
```
wangfenjin, line 104: Could you try add this after the current line?

```
      # https://github.com/sfackler/rust-openssl/issues/1062#issuecomment-489441940
      - run: echo "VCPKGRS_DYNAMIC=1" | Out-File -FilePath $env:GITHUB_ENV -Append
```

If it still not work, we may try https://github.com/sfackler/rust-openssl/issues/1062#issuecomment-555202055

ankrgyl: Added. Let's see 🤞

ankrgyl:
> If it still not work, we may try https://github.com/sfackler/rust-openssl/issues/1062#issuecomment-555202055

Just pushed a change to try this (although not sure I got the syntax correct).

libduckdb-sys/openssl/find_normal.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    let lib = match pkg_config::Config::new().print_system_libs(false).probe("openssl") {
+        Ok(lib) => lib,
+        Err(e) => {
+            println!("run pkg_config fail: {:?}", e);
+            return Ok(());
+        }
+    };
+
+    super::postprocess(&lib.include_paths);
```
wangfenjin, line 222: I think I get it. It's because the function returns nothing but callback to the postprocess function. Which means in our build.rs the `if let` will not work.

Whether we also pass the value back using this postprocess, or we return the path back to the function

libduckdb-sys/build.rs
```diff
@@ -60,11 +102,45 @@ mod build_bundled {
...
+        cpp_files.extend(manifest.base.cpp_files.clone());
+        include_dirs.extend(manifest.base.include_dirs.clone());
+
         let mut cfg = cc::Build::new();
-        cfg.file(format!("{lib_name}/duckdb.cpp"))
-            .cpp(true)
+
+        #[cfg(feature = "httpfs")]
+        {
+            if let Ok((_, openssl_include_dir)) = super::openssl::get_openssl() {
```
wangfenjin, line 119: This one not work

.github/workflows/rust.yaml
```diff
@@ -101,11 +101,26 @@ jobs:
...
+      - run: vcpkg install openssl:x64-windows
```
wangfenjin, line 105: Here should be OK to change back https://github.com/sfackler/rust-openssl/blob/master/.github/workflows/ci.yml#L94

.github/workflows/rust.yaml
```diff
@@ -101,11 +101,26 @@ jobs:
...
+      # https://github.com/sfackler/rust-openssl/issues/1062#issuecomment-489441940
+      - run: echo "VCPKGRS_DYNAMIC=1" | Out-File -FilePath $env:GITHUB_ENV -Append
```
wangfenjin, line 107: delete all the envs here

libduckdb-sys/openssl/find_normal.rs
wangfenjin: They will exit because the save the included file path in the postprocess function. Yes we need the include_dirs so the compiler can find the header file.

through err may not work, should through Ok

wangfenjin: You can go through the find ssl to understand the whole logic

wangfenjin: You can also grant your repo permission to me if you want me to help fix the issue

ankrgyl: That makes sense, although the `Err()` here is not a true error, it's just a trick to short-circuit the execution logic (basically take advantage of the `?` operator). I would suggest that we change the return type of `Err()` to also include the paths, and then return an `Err(...)` here.

ankrgyl: Also just granted you access. Would love your help!

wangfenjin: If return err we need to handle the err carefully. Yes anyway we need to find way to return the include_dirs

ankrgyl: We just need to return the include path in the error (instead of a unit) and extend the match statement in the main code to handle the error case too! I think it should be very straightforward

wangfenjin:
Some follow up:

- [ ] Add tests
- [ ] remove the generated bindings and always generate on the fly
- [ ] Fix the upgrade.sh to use the Python script

nickpresta:
@wangfenjin Thanks for merging this change in. Any change a new release can be cut with these features available?

wangfenjin:
We try to keep the version same with duckdb for easy maintenance, for now maybe you can use master branch, I always keep master branch works

snth:
Thank you @ankrgyl . Going to try this for my own project at the earliest opportunity.

I've been waiting for this for a while and looking at the changes in the PR, I certainly couldn't have done it without you so very grateful! 🙏🙌

PR #144, opened 2023-03-28
feat: make record_batch_to_chunk public

tshauck:
I wonder if you might consider making this function public, or if not you could share a good way to take an arrow record batch and convert it into a DataChunk.

My use case is passing a data chunk from C++ to Rust, then I'd like to use this to copy an arrow record batch to the data chunk.

### DIFF
--- a/src/vtab/arrow.rs
+++ b/src/vtab/arrow.rs
@@ -192,6 +192,13 @@ pub fn to_duckdb_logical_type(data_type: &DataType) -> Result<LogicalType, Box<d
     }
 }
 
+/// Converts a `RecordBatch` to a `DataChunk` in the DuckDB format.
+///
+/// # Arguments
+///
+/// * `batch` - A reference to the `RecordBatch` to be converted to a `DataChunk`.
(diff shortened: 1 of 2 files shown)


wangfenjin:
It’s OK we can make it public.

could you explain more of your use case? This function is convert arrow to data-chunk in rust, which means you will pass data chunk from rust to c++, not from c++ to rust

tshauck:
Yeah, perhaps there's a better way. I have an extension that's written in C++ and for some table functions it's nice to write them in Rust. I have couple examples where I pass the chunk from C++ to rust and fill the chunk with vectors.

This works pretty well, so maybe I should just stick to it, but for complex types (maps, etc) it seemed easier to create the arrow record batch. Happy to go another way if you think there is one.

wangfenjin:
Understand. I'll merge the PR

Please be noted that this function doesn't support all data types, feel free to raise an issue or create PR yourselves if you found anything missing

tshauck:
I probably will be making an attempt, and'll try to open up a PR.

Thanks for this library -- has given me a nice speed-up on a couple things vs how I was doing it.

Issue #148, opened 2023-04-13
Unable to construct a Connection from a raw pointer


Mause:
Heya @wangfenjin,

I'm looking at using the virtual table functionality you integrated from my project (duckdb-extension-framework) and I haven't quite figured out how to fully switch. The missing piece is a way to construct a `Connection` object from a raw pointer in a function called by DuckDB, which is how DuckDB loads extensions.

Something like this would probably suffice:

```rust
pub unsafe fn from_cpp(db: duckdb_database) -> Result<Connection> {
    Ok(Connection {
        db: RefCell::new(unsafe { InnerConnection::new(db, false)? }),
        cache: StatementCache::with_capacity(0),
        path: None,
    })
}
```
which I could then use like this:
```rust
#[no_mangle]
pub unsafe extern "C" fn deltatable_init_rust(db: *mut c_void) {
    init(db.cast()).expect("init failed");
}

unsafe fn init(db: duckdb_database) -> Result<(), Box<dyn Error>> {
    let connection = Connection::from_cpp(db);

    connection.register_table_function::<DeltaFunction>("read_delta")?;
    Ok(())
}
```

I've had a look through the codebase, and can't see a way of doing this without modifying the duckdb-rs code.

Happy to be corrected if I've missed something!

wangfenjin:
I think we need to add a new API for this, I’ll look into that when I got time. Thanks for the input!

wangfenjin:
This issue should have been resolved

PR #149, opened 2023-04-19
build loadable ext

wangfenjin:
Address #148 

Build example: 
`cargo build --example hello-ext --features="vtab bundled"`

There is one issue when load in duckdb cli:
```
./build/debug/duckdb -unsigned

D load "/path/to/libhello_ext.dylib";
/path/to/duckdb/src/catalog/catalog_entry/table_function_catalog_entry.cpp:8:79: runtime error: member access within address 0x7ff7bbe317c8 which does not point to an object of type 'duckdb::CreateFunctionInfo'
0x7ff7bbe317c8: note: object is of type 'duckdb::CreateTableFunctionInfo'
 00 00 00 00  58 63 8e 29 01 00 00 00  19 96 04 00 10 61 00 00  00 00 e3 bb f7 7f 00 00  20 17 e3 bb
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'duckdb::CreateTableFunctionInfo'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /path/to/duckdb/src/catalog/catalog_entry/table_function_catalog_entry.cpp:8:79 in
```

Seems related to https://stackoverflow.com/a/57304113/1203241 , but haven't figure out the solution.

From https://github.com/Mause/duckdb-deltatable-extension/blob/main/CMakeLists.txt , seems by using `build_loadable_extension` may work, but not sure if we have a "cleaner" way to achieve this.

@Mause Any suggestion?


wangfenjin:
If I download the cli from duckdb release page, it works

Issue #166, opened 2023-05-23
[Enhancement] Integration with arrow2 and polars


therealhieu:
Hi,
I want to make some contributions by integrating arrow2 and polars into this crate.
Like:
- query -> arrow2 array -> polars DataFrame
-  polars DataFrame / arrow2 array -> query ->  polars DataFrame

The idea looks like this: https://github.com/therealhieu/rust-duckdb-workaround

If this makes sense, I will contribute a PR

wangfenjin:
Great idea. I had wanted to do this long time ago but haven't got the chance, good you can help!

Let's try split this into multiple PR, I think the first step is to replace all arrow-rs dependency to arrow2

Thanks!

therealhieu:
I will keep the arrow-rs as current, becaucse I think it's doing well and essential.
Instead,  I will add some feature flags for this integration/conversion between arrow-rs and arrow.

Do you agree?

wangfenjin:
Also fine, but I may wondering how you gonna design the API.

From my side, if we still want to keep the arrow-rs, we may have an API like query_arrow2() so that we can avoid to convert from arrow to arrow2, it might be more efficient and more maintainable

wangfenjin:
Anyway, you can try first. Happy to review the code

PR #169, opened 2023-05-29
Feat: Develop query polars

therealhieu:
#166 
I downgraded the arrow-rs to 39 for version compatibilty with arrow2 and polars


Cargo.toml
```diff
@@ -56,13 +51,16 @@ memchr = "2.3"
 uuid = { version = "1.0", optional = true }
 smallvec = "1.6.1"
 cast = { version = "0.3", features = ["std"] }
-arrow = { version = "40", default-features = false, features = ["prettyprint", "ffi"] }
+arrow = { version = "39.0.0", default-features = false, features = ["prettyprint", "ffi"] }
```
wangfenjin, line 54: DependencyBot will complain about this.

Does polars/arrows update their dependency regularly?

Cargo.toml
```diff
@@ -33,13 +33,8 @@ vtab-arrow = ["vtab", "num"]
...
 buildtime_bindgen = ["libduckdb-sys/buildtime_bindgen"]
-modern-full = [
-    "chrono",
-    "serde_json",
-    "url",
-    "r2d2",
-    "uuid",
-]
+modern-full = ["chrono", "serde_json", "url", "r2d2", "uuid", "polars"]
+polars = ["dep:arrow2", "dep:polars", "dep:polars-core"]
```
wangfenjin, line 37: Is the dep: necessary? If not let's delete it

src/lib.rs
```diff
@@ -86,6 +86,8 @@ pub use crate::{
     transaction::{DropBehavior, Savepoint, Transaction, TransactionBehavior},
     types::ToSql,
 };
+#[cfg(feature = "polars")]
+pub use polars_dataframe::{Polars, PolarsDataFrame};
 
 // re-export dependencies from arrow-rs to minimise version maintenance for crate users
 pub use arrow;
```
wangfenjin, line 93: Do you think we also need to re-export polars like we do for arrow?

https://github.com/wangfenjin/duckdb-rs/pull/93

src/lib.rs
```diff
@@ -1260,6 +1264,30 @@ mod test {
         Ok(())
     }
 
+    #[cfg(feature = "polars")]
```
wangfenjin, line 1267: we can move all tests to polars_dataframe.rs so we don't need the cfg here

src/polars_dataframe.rs
```diff
@@ -0,0 +1,71 @@
...
+
+                Series::try_from((name, arrow2_array)).unwrap()
+            })
+            .collect::<Vec<_>>();
+
+        Some(DataFrame::new_no_checks(series_vec))
+    }
+}
+
+/// Polars DataFrame wrapper
```
wangfenjin, line 53: Why we need the wrapper?

wangfenjin:
You can run `cargo test --features "modern-full extensions-full vtab-loadable"` to make sure the tests pass;
Can run `cargo clippy --all-targets --workspace --all-features -- -D warnings -A clippy::redundant-closure` to make clippy pass.

And fmt the code before commit

Cargo.toml
therealhieu: Yes, but there's still a bit delay. 
I see that `arrow` regularly update their version.
If we want to keep the `arrow` updated, we have to make PRs in `arrow2`, `polars` to update their `arrow` version.

Cargo.toml
therealhieu: Yes, it is. 
I set `arrow2`, `polars`, `polars-core` optional. We only turn them on when feature `polars` enabled.

src/lib.rs
therealhieu: Yes. I will add it.

src/lib.rs
therealhieu: Okay. I'll do it

src/polars_dataframe.rs
therealhieu: The result of `query_arrow` is `Iterator of RecordBatch`. 
I want to keep this pattern, so similarly, the result of 'query_polars' is `Iterator of DataFrame` ( 1 RecordBatch is corresponding to 1 DataFrame). Afterall, we need a method to convert `Iterator of DataFrame` into single one `DataFrame`.

It is not possible to `impl FromIterator<DataFrame> for DataFrame` directly. Therefore, there're possibly 4 ways to do this:

1. Creating a wrapper of polars DataFrame like I did.
Usage:
```rust
let df = stmt.query_polars([])?.collect::<PolarsDataFrame>().into_inner();
```

2.  Creating a trait like 'IntoPolarsDataFrame'. 
```rust
trait IntoPolarsDataFrame {
    fn into_polars_dataframe(self) -> DataFrame;
}

impl<T> IntoPolarsDataFrame for T
where
    T: IntoIterator<Item = DataFrame>,
{
    fn into_polars_dataframe(self) -> DataFrame {
        accumulate_dataframes_vertical_unchecked(self)
    }
}
```
Usage:
```rust
let df = stmt.query_polars([])?.into_polars_dataframe();
```
3. Use a function:
```rust
fn into_polars_dataframe<T:    T: IntoIterator<Item = DataFrame>(T) -> DataFrame {
  accumulate_dataframes_vertical_unchecked(self)
}
```
Usage:
```rust
let df = into_polars_dataframe(stmt.query_polars([])?);
```
4. Make `query_polars` return a single DataFrame directly, `Iterator of DataFrame`.

Currently, I think option 2 looks the best

Cargo.toml
wangfenjin: I mean `polars = ["arrow2", "polars", "polars-core"]` will also work? Or is it a better practise to add dep: ?

therealhieu: `polars = ["arrow2", "polars", "polars-core"]` will not work because items in `[ ]` are features, not dependencies.
The use of `dep` is documented here: https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

src/lib.rs
```diff
@@ -1260,6 +1264,30 @@ mod test {
...
+        use polars::prelude::{df, NamedFrom};
+
+        let db = checked_memory_handle();
+        let sql = "BEGIN TRANSACTION;
+                   CREATE TABLE test(t INTEGER);
+                   INSERT INTO test VALUES (1); INSERT INTO test VALUES (2); INSERT INTO test VALUES (3); INSERT INTO test VALUES (4); INSERT INTO test VALUES (5);
+                   END TRANSACTION;";
+        db.execute_batch(sql)?;
+        let mut stmt = db.prepare("select t from test order by t desc")?;
+        let actual = stmt.query_polars([])?.collect::<PolarsDataFrame>().into_inner();
```
wangfenjin, line 1280: ` let actual: Vec<DataFrame> = db.prepare("select t from test order by t")?.query_polars([])?.collect();`

So we don't need `PolarsDataFrame`?

therealhieu: You review [this](https://github.com/wangfenjin/duckdb-rs/pull/169#discussion_r1209659636) then I will modify.

src/polars_dataframe.rs
wangfenjin: Yes I already read this, I mean we have impl the iterator for Polars, and can collect DataFrame from this, so we don't need this wrapper

wangfenjin: You can read `arrow_batch.rs` file again

therealhieu: We can `collect::<Vec<DataFrame>>()` but I believe we should offer a method to convert this `Vec<DataFrame>` into a single `DataFrame`. Therefore, I proposed 4 options as above.

Cargo.toml
wangfenjin: Then why the current calamine or chrono works? they are also dependencies

wangfenjin: But I get it, I think we can add dep here

src/polars_dataframe.rs
wangfenjin: I suggest we don't do this in this crate, user have many ways to combine a Vec.

And when duckdb returns a Vec of DataFrame, it means the result set is big enough and we should steam the data, convert into one might consume a lot of memory and is a bad practise.

therealhieu: That quite makes sense. Converting `Vec<DataFrame>` to `DataFrame` may be beyond this crate's scope. 

In my case, it was not easy to find out the method `accumulate_dataframes_vertical_unchecked`.

Therefore, my thought is that I will the remove the wrapper code but can I put some suggestions about converting `Vec<DataFrame>` to `DataFrame`  in the docstring of the `query_polars` function? I believe this will be helpful for users.

wangfenjin:
Code looks good to me. Only concern is about arrow crate version. Do we have other option than downgrade the version?

therealhieu:
> Code looks good to me. Only concern is about arrow crate version. Do we have other option than downgrade the version?

Yes. There may be 3 options:
- Fork the `arrow2`, `polars2` repo then modify `arrow-rs` dependencies. In `duckdb` crate, point to those fork repo.
- Actively make PRs to update `arrow-rs` deps in `arrow2`, `polars`. I've been doing  this because I am working on a projecet which heavily depends on `duckdb`, `polars`, `arrow2`.
- Add dependabot to `polars`, `arrow2` crate.

wangfenjin:
As polars [re-export arrow ](https://github.com/pola-rs/polars/blob/main/polars/polars-core/src/export.rs#LL3C15-L3C15) in their crate, I think we can change the logic in [here](https://github.com/therealhieu/duckdb-rs/blob/64c5204779db721f24cbf841352b7a072ebc483e/src/polars_dataframe.rs#LL33C7-L33C7) , add a step2() function to use arrow2 as the data structures, so that we can:

1. avoid the overhead of convert arrow to arrow2
2. keep the arrow dependency of this crate up to date
3. add new api like query_arrow2 to return arrow2 data structures to users

Cargo.toml
```diff
@@ -56,13 +51,15 @@ memchr = "2.3"
...
-arrow = { version = "40", default-features = false, features = ["prettyprint", "ffi"] }
+arrow = { version = "39.0.0", default-features = false, features = ["prettyprint", "ffi"] }
 rust_decimal = "1.14"
 strum = { version = "0.24", features = ["derive"] }
 r2d2 = { version = "0.8.9", optional = true }
 calamine = { version = "0.19.1", optional = true }
 num = { version = "0.4", optional = true, default-features = false, features = ["std"] }
 duckdb-loadable-macros = { version = "0.1.0", path="./duckdb-loadable-macros", optional = true }
+polars = { version = "0.30.0", features = ["dtype-full"], optional = true}
+arrow2 = { version = "0.17.2", features = ["arrow"], optional = true }
```
wangfenjin, line 62: if we can use the re-export arrow2 in polars, then we can remove this dependency entry?

therealhieu: https://github.com/pola-rs/polars/blob/4ce40409d1c2a4bbb70289027137262b78720eaa/Cargo.toml#L43

Currently we cannot, because `polars` does not turn on feature `arrow` for `arrow2`.

therealhieu:
> As polars [re-export arrow ](https://github.com/pola-rs/polars/blob/main/polars/polars-core/src/export.rs#LL3C15-L3C15) in their crate, I think we can change the logic in [here](https://github.com/therealhieu/duckdb-rs/blob/64c5204779db721f24cbf841352b7a072ebc483e/src/polars_dataframe.rs#LL33C7-L33C7) , add a step2() function to use arrow2 as the data structures, so that we can:
> 
> 1. avoid the overhead of convert arrow to arrow2
> 2. keep the arrow dependency of this crate up to date
> 3. add new api like query_arrow2 to return arrow2 data structures to users

I am not good at ffi. However, I will give it a try

wangfenjin:
seems the polars need a lot of memory...

therealhieu:
> seems the polars need a lot of memory...

We can try test only feature `polars` to reduce memory consumption. This may work.
I used to face the memory issue with another repo, the solution was using self-hosted runner :).

wangfenjin:
Haha self-hosted runner cost money

Yeah a good suggestion to split the testing, for now I just disable the polar testing in windows

wangfenjin:
polars increase about 30% CI time 😭

therealhieu:
Sorry I'm kinda busy this week.
I will try finish this PR this weekend.

therealhieu:
Got this error in the CI when running `cargo clippy --all-targets --workspace --all-features -- -D warnings -A clippy::redundant-closure`
```
error: redundant clone
Error:   --> duckdb-loadable-macros/src/lib.rs:14:46
   |
14 |             let c_entrypoint = func.sig.ident.clone();
   |                                              ^^^^^^^^ help: remove this
   |
note: cloned value is neither consumed nor mutated
  --> duckdb-loadable-macros/src/lib.rs:14:32
   |
14 |             let c_entrypoint = func.sig.ident.clone();
   |                                ^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
   = note: `-D clippy::redundant-clone` implied by `-D warnings`

error: redundant clone
Error:   --> duckdb-loadable-macros/src/lib.rs:19:60
   |
19 |             let prefixed_original_function = func.sig.ident.clone();
   |                                                            ^^^^^^^^ help: remove this
   |
note: cloned value is neither consumed nor mutated
  --> duckdb-loadable-macros/src/lib.rs:19:46
   |
19 |             let prefixed_original_function = func.sig.ident.clone();
   |                                              ^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone

error: could not compile `duckdb-loadable-macros` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

However, in my local, clippy linting does not result in any errors in my local. Removing those clone() gives me compile time errors.

wangfenjin:
You can merge main branch, I had fix that in main

therealhieu:
is this branch ready to be merged?

wangfenjin:
Should be ready now! I’ll review the code again later today. Thanks again for the effort!

therealhieu:
it's my pleasure!

therealhieu:
Can you make a new release? I currently need this feature.

anndev4:
Hello @wangfenjin  i also need this feature on my side, can you release please ?

wangfenjin:
@therealhieu @anndev4 just now tried that, but found it's not easy to release a new version. As we are currently trying to stay with the duckdb repo for version release, and some part depends on the version, if we make the change, we need to change several part.

Maybe you guys can use the master branch, I always keep the master branch works normally.

PR #238, opened 2023-11-23
Add method to pass any setting to DuckDB config

mlafeldt:
Most settings from https://duckdb.org/docs/sql/configuration.html currently can't be configured since the `set` method is private. This PR adds a new `with` method that makes it easy to compose a config with arbitrary settings.


Mytherin:
Thanks for the PR! Could we add a test with an invalid setting as well?

mlafeldt:
> Thanks for the PR! Could we add a test with an invalid setting as well?

Absolutely! Done.

Mytherin:
Thanks!

PR #259, opened 2024-02-08
Make the ArrowVTab module public

phillipleblanc:
I assume from the comment in [this issue](https://github.com/duckdb/duckdb-rs/issues/79#issuecomment-1479265668) that the ArrowVTab module was intended to be made public. This PR makes it public so that consumers can actually use it.

I've tested this code in my fork and it works great.


Mause (approved):

Maxxen (approved):

Mytherin:
Thanks!

Issue #286, opened 2024-04-08
BUG: Panic on `SELECT * FROM arrow(?, ?)` when using `TimestampMillisecondArray` arrow type.


Jeadie:
## Overview
When creating a table in DuckDB using an Arrow record batch, the duckDB panics when the RecordBatch has a column of type `TimestampMillisecondArray` (and true of other timestamp array types).

#### main.rs
```rust
use std::sync::Arc;
use duckdb::{arrow::{array::{ArrayRef, Int32Array, StringArray, TimestampMillisecondArray}, record_batch::RecordBatch}, vtab::arrow::{arrow_recordbatch_to_query_params, ArrowVTab}, Connection};

fn main() {
    let a: ArrayRef = Arc::new(Int32Array::from(vec![1, 2, 3]));
    let b: ArrayRef = Arc::new(StringArray::from(vec!["a", "b", "a"]));
    let c: ArrayRef = Arc::new(TimestampMillisecondArray::from(vec![11111111, 22222222, 33333333]).with_timezone("+10:00".to_string()));
    let batch = RecordBatch::try_from_iter(vec![
        ("x", a),
        ("y", b),
        ("z", c),
    ]).unwrap();

    let db = Connection::open_in_memory().unwrap();
    db.register_table_function::<ArrowVTab>("arrow").unwrap();
    let name = "my_table";  

    let mut stmt = db.prepare(format!(r#"CREATE TABLE "{name}" AS SELECT * FROM arrow(?, ?)"#).as_str()).unwrap();
    stmt.execute(arrow_recordbatch_to_query_params(batch)).unwrap();
    
    match db.query_row("SELECT count(1) FROM my_table", [], |x| x.get::<_, i32>(0),) {
        Ok(x) => println!("num_rows: {}", x),
        Err(e) => println!("Error: {}", e),
    }
}
```

#### `cargo.toml`
```toml
[package]
name = "example"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
arrow = "51.0.0"
duckdb = {path="../", features=["bundled",
    "r2d2",
    "vtab",
    "vtab-arrow"] 
}
```

#### Stacktrace
```shell
thread 'main' panicked at src/vtab/arrow.rs:334:13:
not yet implemented
stack backtrace:
   0: rust_begin_unwind
[...]

PR #289, opened 2024-04-10
Support more time types to arrow vtab

Maxxen:
Spent a ton of time creating conversion [following the behavior in the c++ code](https://github.com/duckdb/duckdb/blob/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/function/table/arrow_conversion.cpp#L751-L773) until I realized the arrow crate already has array cast functions built in.

I've left out intervals and durations for now as they have more complicated conversion logic and intervals are broken anyway. (see https://github.com/duckdb/duckdb-wasm/issues/1696#issuecomment-2047272977). 

This should fix: #286

### DIFF
--- a/src/vtab/arrow.rs
+++ b/src/vtab/arrow.rs
@@ -609,7 +609,11 @@ mod test {
         let rb = stmt.query_arrow(param)?.next().expect("no record batch");
 
         let output_any_array = rb.column(0);
-        assert_eq!(output_any_array.data_type(), expected_array.data_type());
+        match (output_any_array.data_type(), expected_array.data_type()) {
+            // TODO: DuckDB doesnt return timestamp_tz properly yet, so we just check that the units are the same
+            (DataType::Timestamp(unit_a, _), DataType::Timestamp(unit_b, _)) => assert_eq!(unit_a, unit_b),
+            (a, b) => assert_eq!(a, b),
+        }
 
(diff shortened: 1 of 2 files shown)


phillipleblanc (approved):

src/vtab/arrow.rs
```diff
@@ -324,22 +336,35 @@ fn primitive_array_to_vector(array: &dyn Array, out: &mut dyn Vector) {
...
-        // DataType::Decimal256(_, _) => {
-        //     primitive_array_to_flat_vector::<Decimal256Type>(
-        //         as_primitive_array(array),
-        //         out.as_mut_any().downcast_mut().unwrap(),
-        //     );
-        // }
-        _ => {
-            todo!()
+        DataType::Decimal256(_, _) => todo!("Decimal256 is not supported yet"),
+        DataType::Timestamp(_, tz) => primitive_array_to_flat_vector_cast::<TimestampMicrosecondType>(
```
Swoorup, line 340: shouldn't the timeunit field be respected? How would this support data types such as `TIMESTAMP_NS`

Maxxen: Ah, youre right, its only the TIMESTAMPTZ type that only supports microsecond precision (and needs to be scaled). Ill push a fix soon

PR #292, opened 2024-04-13
feat: add support for reading lists

Mause:
@Swoorup if you have a chance to review this design, I'd appreciate it as well

### DIFF
--- a/src/types/value_ref.rs
+++ b/src/types/value_ref.rs
@@ -95,8 +95,9 @@ impl ValueRef<'_> {
         }
     }
 
+    /// Returns an owned version of this ValueRef
     pub fn to_owned(&self) -> Value {
-        (*self).clone().into()
+        (*self).into()
     }
 }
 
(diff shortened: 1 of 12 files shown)


src/types/value_ref.rs
```diff
@@ -87,8 +92,14 @@ impl ValueRef<'_> {
             ValueRef::Blob(_) => Type::Blob,
             ValueRef::Date32(_) => Type::Date32,
             ValueRef::Time64(..) => Type::Time64,
+            ValueRef::List(arr, _) => arr.data_type().into(),
         }
     }
+
+    /// Returns an owned version of this ValueRef
+    pub fn to_owned(&self) -> Value {
```
Swoorup, line 100: Could instead using the standard `ToOwned` trait impl here.
```rust
impl ToOwned for ValueRef<'_> {
    type Owned = Value;

    fn to_owned(&self) -> Self::Owned {
      *self.into()
    }
}
```

Swoorup (approved):
Looks fine by me.

src/types/value_ref.rs
Mause: Unfortunately this conflicts with the default implementation provided by Clone, and would require reworking elsewhere

PR #295, opened 2024-04-17
add vtab and chrono modules to docs

Mause:


Cargo.toml
```diff
@@ -85,7 +85,7 @@ path = "libduckdb-sys"
 version = "0.10.1"
 
 [package.metadata.docs.rs]
-features = []
+features = ['vtab', 'chrono']
```
Maxxen, line 88: Did you mean to commit this?

Mause: Yes, it's what adds these modules to docs.rs

Mause: They're presently missing from here: https://docs.rs/duckdb/latest/duckdb/#modules

Maxxen: Cool thanks

PR #297, opened 2024-04-22
feat: enum read support

Mause:


src/types/value_ref.rs
```diff
@@ -213,6 +246,7 @@ impl<'a> From<&'a Value> for ValueRef<'a> {
             Value::Time64(t, d) => ValueRef::Time64(t, d),
             Value::Interval { months, days, nanos } => ValueRef::Interval { months, days, nanos },
             Value::List(..) => unimplemented!(),
+            Value::Enum(..) => todo!(),
         }
```
Maxxen, line 250: We should maybe change this to a TryFrom instead in the future

Mause: That's a good idea yeah

PR #316, opened 2024-05-24
Fixes the issue ignoring nulls when copy from arrow array to flat vector 

### LINKED ISSUE #157: Dealing with arrow nulls?
Hello!

I'm trying to copy an arrow record batch into a duckdb chunk via `record_batch_to_duckdb_data_chunk`. Generally things are working however it seems that items that should be null aren't working properly. I think on downcasting or otherwise, things are getting dropped, e.g. a null i32 becomes 0. I'm only tried the basic primitive types.

I've created a little test here: https://github.com/tshauck/duckdb-rs/commit/0d3b87b59a96189f2a6701681cca2e3bf1874a87 (I think `validity` shouldn't be null).

I'm happy to try to take a look, but any pointers where to start would be helpful.

y-f-u:
This fixes #157 

When copying an arrow array which contains null into duckdb, the value is converted to default value before the fix.

The fix loops through the null marks in the array and call vector `set_null` accordingly.


Maxxen:
Thanks! LGTM

Issue #320, opened 2024-05-30
using vendored_openssl with bundled ducdkb-rs 


muralisoundararajan:
I am trying to build duckdb-rs for AWS Lambda Runtime with vendored_openssl approach.

Is there a way to pass 'vendored' feature information from libducksys crate to openssl module within the libducksys crate.

muralisoundararajan:
@Maxxen Would it be feasible to include  https://github.com/duckdb/duckdb-rs/pull/321https://github.com/duckdb/duckdb-rs/pull/321 in coming release if the changes are acceptable?

Maxxen:
Sure, ill have a look

PR #321, opened 2024-05-31
Add ability to pass vendored feature to openssl in libduckdb-sys

muralisoundararajan:
openssl module imported from [rust-openssl ](https://github.com/Mause/duckdb-rs/blob/struct/libduckdb-sys/openssl/mod.rs#L2) provides an ability to use vendored openssl as a feature.  In libduckdb-sys, when it is introduced as module, the ability to provide "vendored" feature was not added to libduckdb-sys and duckdb  crates.

This change adds ability to pass "vendored" feature to "openssl" module within the "libduckdb-sys"

Resolves https://github.com/duckdb/duckdb-rs/issues/320


Mause (approved):
Looks good to me! Appreciate the renamed features to include what they really do as well

muralisoundararajan:
@Mause Thanks for the review.

CI pipeline is failing with the same reason as main line due to new version of [OpenSSL 3.3.1](https://slproweb.com/download/Win64OpenSSL-3_3_1.exe) available in choco and runner-images for Windows is using the previous version.

By any chance, would it be possible to wait for this week duckdb release to wait till we have this issue is resolved, so that this PR can be part of the release?

Maxxen:
@muralisoundararajan Its not obvious to me how to solve this issue (googling for this gives me no results). Could you share some more information from your investigation? Is this just a problem with choco that we have to wait for them to fix or can we adjust something in our CI pipeline to resolve it? 

I suspect we can maybe remove the choco openssl install completely and just rely on the openssl version provided by the runner environment?

Maxxen:
Actually, I think I fixed it in #327

muralisoundararajan:
@Maxxen Yes, that seem to resolve the issue. Thanks for the quick turnaround. 

I rebased my changes, would it be possible to run the CI workflow.

muralisoundararajan:
Another option I was contemplating was to use openssl_vendored in msvc environment during the failure.

Maxxen:
Thanks! Sorry for the friction :)

PR #332, opened 2024-06-07
upgrade arrow-rs to 52

Mause:


Maxxen:
Thanks! Great that the time issue is fixed

PR #336, opened 2024-06-10
Update duckdb-rs to DuckDB v1.0.0

Mytherin:
(Hopefully) includes a work-around for the broken github actions Windows runners (see https://github.com/actions/runner-images/issues/10004)

Issue #211, opened 2023-09-22
Segmentation fault when using appender API


jubos:
When using the appender API, you can generate a segfault if you break a foreign key constraint.

If you use the insert API, you will get a returned error when you conduct an insert that breaks a foreign key constraint.

The reproduction of the bug is in this repo: https://github.com/jubos/duckdb-sandbox

You can run the following to generate it:
```
cargo run test.db
```

wangfenjin:
could you try reproduce this issue using the c/c++ API? If reproducible means we should report it in duckdb repo

ctsrc:
Had a similar thing happen to me when I appended rows with Primary Key values that were already in the table. Instead of returning an error it segfaulted.

Platform is macOS

samuelcolvin:
I'm getting the same, @wangfenjin @Mause, since @jubos has provided a reproduction of the issue, surely this deserves to be investigated as a priority?

Mause:
> I'm getting the same, @wangfenjin @Mause, since @jubos has provided a reproduction of the issue, surely this deserves to be investigated as a priority?

I've investigated it yeah, but as neither myself nor Wang have merge access, our ability to fix it is limited

samuelcolvin:
Sorry to ping you, I assumed since you're the most active contributors, you would have merge access.

To be honest, I think #287 is more problematic as that's (afaik) correct usage causing a memory error.

Swoorup:
I have a branch where I've fixed quite a lot of bugs related to complex/nested types but I am waiting few of my changes to be merged before I can proceed. For now I just have a local copy.

samuelcolvin:
That's great news @Swoorup, once you have a PR that you think might fix this or #287, let me know and I can test it with my code.

I'm very keen to use duckdb from Rust lots more, but I'd really like the memory errors to get fixed. We (Pydantic, the company) might be able to sponsor significantly if anyone is willing and able to work through these issues.

Swoorup:
I roughly tested this particular issue in my branch and I get, which I assume is correct?

```shell
Failed to flush appender: DuckDBFailure(Error { code: Unknown, extended_code: 1 }, Some("Violates foreign key constraint because key \"id: 0\" does not exist in the referenced table"))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The issue with generated columns is different though I think. It doesn't segfault but panics on mismatched schema, as my approach is first getting the table schema and validating it against what will be appended. 

[...]

Issue #188, opened 2023-07-08
he memory panic 


web3creator:
[    pub fn copy<T: Copy>(&mut self, data: &[T]) {](https://github.com/wangfenjin/duckdb-rs/blob/ae8b036f855e60c50dfef14e12a189d912f6dd52/src/vtab/vector.rs#L85)

Regarding the memory panic issue, if the size of my data exceeds 1k, it will panic here. Is there any solution?

wangfenjin:
May related to the vector capacity, need to check the details

Issue #81, opened 2022-09-10
How to insert lists, maps, and structs?


Timmmm:
Maybe I missed it, but I couldn't see how to insert a row if one of the columns is e.g. `TEXT[]`. I naively tried `params![a_vec_of_str]` but `ToSql` is only implemented for `Vec<u8>` (I guess for blobs).

The DuckDB docs aren't super clear but it looks like you're supposed to use `list_create('a', 'b', 'c')` but I'm not sure how to do that with `params![]`.

wangfenjin:
Haven’t been supported in rust yet

digizeph:
@wangfenjin appreciate your work! Just want to insert here that missing insert of lists is a deal breaker for my projects. Hope to see that implemented soon. Cheers!

Swoorup:
Recent support for Union types too. 👍🏼 Natural fits with Rust's enum.

wangfenjin:
@neverchanje it'll be great if you can help. Thanks

Swoorup:
Would be great if we could also be able to retrieve SRUCT, MAP, ENUMs and UNION along with being able to insert them. I think the insert just needs string representation vs the retrieval needs to be convert from arrow format? 

Am I correct @wangfenjin ?

wangfenjin:
Correct. 

1. If you want to support bind struct or other type as params, you may need to update [bind_params](https://github.com/wangfenjin/duckdb-rs/blob/d2f611f922792ec6943a4f65301c81e247864fc0/src/statement.rs#L414)
2. If you want to support parse sql result as rust native datatypes like map, you may need to update [get](https://github.com/wangfenjin/duckdb-rs/blob/d2f611f922792ec6943a4f65301c81e247864fc0/src/row.rs#L544)
3. If it's a new native datatypes, you need to impl the FromSql and ToSql trait, example [serde_json](https://github.com/wangfenjin/duckdb-rs/blob/d2f611f922792ec6943a4f65301c81e247864fc0/src/types/serde_json.rs#L8-L27)
4. You may also want to add new type to enum [Value](https://github.com/wangfenjin/duckdb-rs/blob/d2f611f922792ec6943a4f65301c81e247864fc0/src/types/value.rs#L10) when necessary, if you do that compiler will teach you multiple related places that also need to updated

@Swoorup

wangfenjin:
The overall idea is in order to make it easier for users, we need to implement the convert logic of from/to sql

[...]

PR #337, opened 2024-06-11
Duckdb updates

Swoorup:
1. Strong types for duck db vectors.
2. Ensure correct size are reserved before writing to duckdb vectors.
3. Introduce `DataChunkHandle` for manipulating data chunks, and wrap unsafe functions at bay. There are still few handle like `duckdb_appender` that is better off being wrap instead of sprinkling unsafe everywhere.
4. Add union, struct type and nested type without the duplicated code. (It should be simple enough to now add other nested types like `Map`)
5. Appender keeps a cached copy of the schema for error handling and insertions.
6. Fix large amount of record insertion (Respect the duckdb limit)
7. Remove unsafe code for traits that need not be unsafe. Entire idea for rust is to limit usage of unsafe code when providing as a library.
8. For unit testing, I also added [arrow-convert](https://github.com/Swoorup/arrow-convert) library which makes it easier to derive arrow arrays directly from rust specific unions/structs than hand writing it.


Swoorup:
Fixes or partially fixes the following:

- https://github.com/duckdb/duckdb-rs/issues/211
- https://github.com/duckdb/duckdb-rs/issues/188
- https://github.com/duckdb/duckdb-rs/issues/81

Swoorup:
Union feature is blocked by a suitable release i.e which includes the fix https://github.com/duckdb/duckdb/pull/11326 https://github.com/duckdb/duckdb-rs/pull/336

Mytherin:
Thanks for the PR! Could you perhaps split this up into smaller chunks so that it is easier to review?

Swoorup:
**NOTICE**: I have abandoned this work. Anyone is welcome to take over and integrate bits and pieces they may find useful.

PR #339, opened 2024-06-13
Remove mut from raw_query

rijkvp:
The `&mut` reference was not required here, this solved a use case for us where we wanted to query while the statement was immutably borrowed.


Mytherin:
Thanks!

PR #343, opened 2024-06-20
Allow `Connection` to be interrupted

### LINKED ISSUE #342: Cannot use `duckdb_interrupt` due to private `InnerConnection`
The `Connection` hides away the raw duckdb connection object in private fields:
```rust
/// A connection to a DuckDB database.
pub struct Connection {
    db: RefCell<InnerConnection>,
    cache: StatementCache,
    path: Option<PathBuf>,
}
```
`InnerConnection` contains a `ffi::duckdb_connection`...
```rust
pub struct InnerConnection {
    pub db: ffi::duckdb_database,
    pub con: ffi::duckdb_connection,
    owned: bool,
}
```

...which is required to call `ffi::duckdb_interrupt`:
```rust
extern "C" {
    #[doc = "Interrupt running query\n\n connection: The connection to interrupt"]
    pub fn duckdb_interrupt(connection: duckdb_connection);
}
``` 

There's no way to call `duckdb_interrupt` (and other `ffi` functions like it) with a `Connection`, as we can't access its fields. Exposing a `Connection::interrupt()` method would be one way around this.

I'm happy to make the contribution if the maintainers can advise on how you would like this implemented.

rebasedming:
Closes #342 

Allows the `Connection` to be interrupted, which is useful for cancelling long-running queries.


Mause (changes requested):
Can we please get some tests and docs added? And remove the dsstore files

rebasedming:
> Can we please get some tests and docs added? And remove the dsstore files

Yep! Put this up hastily to see if there was appetite for this functionality. I can clean it up now and add tests if there is.

Mytherin:
Could you merge this with main, and perhaps also add some tests?

rebasedming:
Yes, I got distracted with other things the last few weeks. I'll finish up this PR.

rebasedming:
Hey @Mause @Mytherin, I could use your input on how you'd like to see this tested. What I'm envisioning is spawning a long-running query in one thread, and then interrupting the same connection from another thread. Unfortunately, once a `Connection` is moved to a thread, it can't really be called from another thread. For instance, the following code won't compile because `db` is moved to the thread and you can't call `interrupt` on it afterward.

```rust
  #[test]
  fn test_interrupt() -> Result<()> {
      let db = checked_memory_handle();

      let query_handle = std::thread::spawn(move || {
          let mut stmt = db.prepare("CREATE TABLE foo AS SELECT generate_series(1, 10000000) AS id;")?;
          stmt.execute([])?;
          Ok(())
      });

      // Wait for the query to start
      std::thread::sleep(std::time::Duration::from_millis(1000));
      db.interrupt();

      let query_result: Result<(), Error> = query_handle.join().unwrap();
      assert!(query_result.is_err());
      Ok(())
  }
```

Mytherin:
That idea for a test sounds good to me. 

> Unfortunately, once a Connection is moved to a thread, it can't really be called from another thread. For instance, the following code won't compile because db is moved to the thread and you can't call interrupt on it afterward.

I'm not too familiar with Rust specifics - but how do you envision this method to be used in practice if this is not possible? Maybe there needs to be some annotation that signifies it is safe to call `interrupt` from a different thread?

rebasedming:
> That idea for a test sounds good to me.
> 
> > Unfortunately, once a Connection is moved to a thread, it can't really be called from another thread. For instance, the following code won't compile because db is moved to the thread and you can't call interrupt on it afterward.
> 
> I'm not too familiar with Rust specifics - but how do you envision this method to be used in practice if this is not possible? Maybe there needs to be some annotation that signifies it is safe to call `interrupt` from a different thread?

In practice, I've stored the `Connection` in a [static global](https://github.com/paradedb/paradedb/blob/079350071a1280cddcf10fa1359ce63fcc704c5b/pg_lakehouse/src/duckdb/connection.rs#L35). I do the same for the tests if you're comfortable with it?

Mytherin:
If the tests are completely stand-alone I don't mind - but ideally we don't ship any static globals. Is there no other way?

mlafeldt:
Superseded by #493

PR #344, opened 2024-06-21
Set nulls correctly for all type of arrays/vectors

y-f-u:
This sets the correct nulls for strings, struct, list, array types when copying data from arrow array to duckdb vectors.

When inserting nulls to columns with these types, the null will be correctly set instead of the default value like: `""`, `{k: ""}`, etc.


crates/duckdb/src/vtab/vector.rs
```diff
@@ -279,4 +295,13 @@ impl StructVector {
         let logical_type = self.logical_type();
         unsafe { duckdb_struct_type_child_count(logical_type.ptr) as usize }
     }
+
```
Swoorup, line 298: Some of these makes no sense. The struct vector doesn't store the columns directly. Think that goes for all Arrays, Union,

y-f-u: Thanks for the comment. This change is about to set the validity on the vector, so duckdb knows when to set null correctly.

See added tests.

Mytherin:
Thanks for the PR! LGTM - could you just merge with main so the CI can rerun?

Mytherin:
Thanks - looks like the CI now fails on the newly added test. Could you have a look when you have time?

y-f-u:
yeah, will do. Sorry got carried away by something else, looks like an easy fix.

y-f-u:
that's still something wrong with invalid cfg feature for time. Which is not introduced by this branch, I took a shot on fixing that but not really sure how the `time` features come from and how it got used.

Mytherin:
Thanks for looking into it - could you merge with main again? The CI should be fixed again in https://github.com/duckdb/duckdb-rs/pull/375

Mytherin:
Thanks!

PR #348, opened 2024-06-23
feat: add bindings for Scalar function API

rustyconover:
Add some initial bindings for the scalar function C API.


Mytherin:
Thanks for the PR! Could you rebase with main?

crates/libduckdb-sys/src/lib.rs
```diff
@@ -14,6 +14,24 @@ pub use bindings::*;
 pub const DuckDBError: duckdb_state = duckdb_state_DuckDBError;
 pub const DuckDBSuccess: duckdb_state = duckdb_state_DuckDBSuccess;
 
+use std::slice;
+impl From<&duckdb_string_t> for String {
```
Swoorup, line 18: `DuckDbString` from https://github.com/duckdb/duckdb-rs/pull/361 might be beneficial here, instead of preemptively converting to Rust String

PR #359, opened 2024-07-19
Move files out of vtab to core module

Swoorup:
As a prep for breaking down https://github.com/duckdb/duckdb-rs/pull/337

Add few rustfmt configs, for enforcing consistency.  Commented the ones that currently require night but can be enforced later. 

This is a refactoring only changes that moves files out of vtab to core.


Mytherin:
Thanks!

PR #360, opened 2024-07-19
feat: set duckdb api to rust and add custom_user_agent config

hrl20:
- this is so we can distinguish rust from capi.
- caller can configure the custom_user_agent duckdb config.


Mytherin:
Thanks! LGTM - could you just run the formatter?

Mytherin:
Thanks!

PR #361, opened 2024-07-20
Rename `LogicalType` and `DataChunk` to have `Handle` suffix

Swoorup:
More refactoring changes from https://github.com/duckdb/duckdb-rs/pull/337

- Rename `LogicalType` and `DataChunk` to have `Handle` suffix, as these are just reference handle to the duckdb'c api's. `LogicalType` and `DataChunk` will be used later to contain more information about the themselves.
- Introduce `DuckDbString` as this need to be cleanup using a Duckdb C api. Mostly to wrap unsafe functions.
- Use `num-integer` version from root manifest toml for consistency
- Also mark functions/convert to unsafe equivalent instead of using the from trait, when getting their inner field directly from c api, to mark it as unsafe.


Mytherin:
Thanks!

PR #363, opened 2024-07-23
add thread local data initialization

0xcaff:
This allows for thread local initialization of multi-threaded vtable execution. I've validated this works by pulling this patch into a vendored copy of this library in my project (https://github.com/0xcaff/duckdb_protobuf/commits/threads/)

### DIFF
--- a/crates/duckdb/src/vtab/mod.rs
+++ b/crates/duckdb/src/vtab/mod.rs
@@ -143,6 +143,19 @@ where
     }
 }
 
+unsafe extern "C" fn local_init<T>(info: duckdb_bind_info)
+where
+    T: VTabLocalData,
+{
+    let info = InitInfo::from(info);
+    let data = malloc_data_c::<T::LocalInitData>();
+    let result = T::local_init(&info, data);
+    info.set_init_data(data.cast(), Some(drop_data_c::<T::LocalInitData>));
+    if result.is_err() {
+        info.set_error(&result.err().unwrap().to_string());
+    }
+}
+
 unsafe extern "C" fn bind<T>(info: duckdb_bind_info)
 where
     T: VTab,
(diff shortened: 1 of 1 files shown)


crates/duckdb/src/vtab/mod.rs
```diff
@@ -160,23 +173,61 @@ impl Connection {
...
+    /// See [`InitInfo::set_max_threads`] for enabling multithreaded execution.
+    ///
+    /// This function is unsafe because it performs raw pointer dereferencing on the `data` argument.
+    /// The caller is responsible for ensuring that:
+    ///
+    /// - The `data` pointer is non-null and points to a valid `LocalInitData` instance.
+    /// - There is no data race when accessing `data`, meaning if `data` is accessed from multiple threads,
+    ///   proper synchronization is required.
+    /// - The lifetime of `data` extends beyond the scope of this call to avoid use-after-free errors.
+    unsafe fn local_init(init: &InitInfo, data: *mut Self::LocalInitData) -> Result<(), Box<dyn std::error::Error>>;
```
Swoorup, line 222: Is the unsafe here necessary? The implementation block could contain unsafe without the trait function being unsafe?

0xcaff: The init, bind and func functions were recently changed to unsafe https://github.com/duckdb/duckdb-rs/commit/ff7d11db7098316efb588bce61406327dae57cc5 This is simply following the same pattern as the others.

As for whether it needs to be unsafe, I believe it is because the way memory is allocated for the data functions doesn't correctly zero initialize the memory https://github.com/duckdb/duckdb-rs/blob/1c73aef679d0a33f380861c8061c82bf6edbda07/crates/duckdb/src/vtab/mod.rs#L138

I get around this by storing a pointer in the data field and allocating a Box manually inside the init functions. This is something worth improving (use Box to allocate with rust semantics instead of malloc and pass a reference) but I'll leave that for a separate PR.

0xcaff:
closing due to inactivity. decided to vendor the changes into my project for now.

PR #375, opened 2024-08-27
Fix CI

samansmink:
@Mytherin I think this solves the CI issues, likely something changed upstream making clippy more strict. 

I've removed the `time` feature completely. It doesn't appear to be used anywhere and seems like a leftover from copying rusqlite stuff. Also remove the large comment which is just copied from rusqlite and is just confusing overall: `DuckDB uses a [dynamic type system](https://www.sqlite.org/datatype3.html)` 

I just wanted to restore CI here quickly, but i think we should look into more thoroughly trimming dead features/code from the package. I feel there's a lot of stuff that is not really used and could potentially be removed to make the package easier to understand.


Swoorup:
You could also remove https://github.com/duckdb/duckdb-rs/blob/main/crates/duckdb/src/util/small_cstr.rs As its not used anywhere.

Mytherin:
Thanks!

Issue #370, opened 2024-08-08
Rust extensions based on Extension C API


samansmink:
Hi 👋 

I've been working on the new C API for extensions in duckdb/duckdb (see https://github.com/duckdb/duckdb/pull/12682). TLDR: we've added a header to duckdb which contains basically a struct of function pointers with most of the C API. This struct is then passed to a loadable extension on load.  This is basically the same mechanism that [sqlite has](https://github.com/sqlite/sqlite/blob/master/src/sqlite3ext.h). 

### Goals
One of the goals of this API is to be able to write clean pure-rust extensions for DuckDB that can be used across multiple duckdb versions. Basically I want an extension template like `https://github.com/duckdb/extension-template`, but than in pure-rust based on this new C API. 

Now that that PR is merged, I would like to get started on this.

### Approach
Good for me, rusqlite, the repo this one is based on, is way ahead of us and already has support for sqlite's variant of the extension api struct since https://github.com/rusqlite/rusqlite/pull/1362.

So in theory, all we need to do is copy (the parts we like) of rusqlite for this to generate the loadable extensions. Then tie everything together for the 

### The catch
The main catch is my Rust knowledge is pretty minimal. Therefore this will also be a bit of a rust-learning-experience for me. However, given the fact that it's mostly a copy-and-paste exercise from rusqlite, this should be fine? ;)

finally, any help or advice is very welcome!

samansmink:
Hey @0xcaff, That's great to hear! Lets setup a quick call I would love to get your take on this. I will drop you an email.

I have also fiddled with it and I have a (very messy) PoC working here: https://github.com/samansmink/duckdb-rs/tree/poc-rust-c-extension-api. I think it looks pretty similar to your direction, but I also added the attribute macro for the entrypoint, handled the version properly and added a demo function to the examples which uses the new entrypoint. I actually managed to load the extension from DuckDB too! :)  

So basically the end-to-end is there, now its a matter of thinking about how to get things integrated cleanly. One of the things that I find a little awkard about the way things are setup in my feature branch now is that the libduckdbsys package is now basically fundamentally different based on this `loadable_extension` feature, making it not really a feature but almost transforming it completely into a different package.

0xcaff:
Nice! It looks like my changes are pretty similar to yours. I think the open questions from me are

* Versioning. It seems like you allow specifying a min-version in your proc macro but it seems unused.
* Bundled build. Neither of our implementations support using a bundled version of the bindings.
[...]

PR #381, opened 2024-09-16
C Extension API 

samansmink:
This PR makes the first step towards proper support for https://github.com/duckdb/duckdb-rs/issues/370.
In the process it also bumps duckdb-rs to duckdb v1.1.1

With this PR I we introduce the experimental `duckdb_entrypoint_c_api` proc macro that will basically handle everything for you as can be seen in this snippet from `crates/duckdb/examples/hello-ext-capi`:

```rust
#[duckdb_entrypoint_c_api(ext_name = "rusty_quack", min_duckdb_version = "v0.0.1")]
pub unsafe fn ExtensionEntrypoint(con: Connection) -> Result<(), Box<dyn Error>> {
    con.register_table_function::<HelloVTab>("hello").expect("Failed to register hello table function");
    Ok(())
}
```

Note that this relies on the new extension C api added in https://github.com/duckdb/duckdb/pull/12682.

The main idea is that these binaries are (going to be) forwards compatible with new duckdb releases. This compatibility can be configured using the `min_duckdb_version` parament of the proc macro. For now, this version is set to a separate C extension API version (which is v0.0.1) in current duckdb stable. However in the future, this version will likely be simply the duckdb version. This means that loadable extensions currently produced with the `duckdb_entrypoint_c_api` proc are **not** stable right now. This stability *is* expected to be possible in DuckDB v1.2.0 in a few months.

Finally, the extension binaries in this PR are not ready to be loaded by DuckDB directly, they need to have their extension metadata appended. I will add this script to the https://github.com/duckdb/extension-ci-tools repo and create a rust extension template that uses the script to perform the final transform step for rust shared libary to loadable duckdb extension

### follow ups
- testing: the upgrade script running the bindgen performs no tests for the pre-generated bindings currently 
- create the demo extension template


0xcaff:
nice nice! I'll try to get this reviewed today evening or early tomorrow

crates/libduckdb-sys/upgrade.sh
```diff
@@ -1,33 +1,45 @@
...
+cd "$SCRIPT_DIR"
 cargo clean
 mkdir -p "$SCRIPT_DIR/../../target" "$SCRIPT_DIR/duckdb"
 export DUCKDB_LIB_DIR="$SCRIPT_DIR/duckdb"
-export DU_INCLUDE_DIR="$DUCKDB_LIB_DIR"
 
 # Download and extract amalgamation
-DUCKDB_VERSION=v1.0.0
+# todo: fix this version
+DUCKDB_VERSION=v1.1.0
```
0xcaff, line 16: we can remove the todo now

crates/libduckdb-sys/upgrade.sh
```diff
@@ -1,33 +1,45 @@
...
 
+# Regenerate bindgen file for DUCKDB
+cd "$SCRIPT_DIR"
+rm -f "$SCRIPT_DIR/src/bindgen_bundled_version_loadable.rs"
+find "$SCRIPT_DIR/../../target" -type f -name bindgen.rs -exec rm {} \;
+cargo build --features "extensions-full buildtime_bindgen loadable_extension"
+find "$SCRIPT_DIR/../../target" -type f -name bindgen.rs -exec cp {} "$SCRIPT_DIR/src/bindgen_bundled_version_loadable.rs" \;
+
+# Sanity checks
+# FIXME: how to test this here?
```
0xcaff, line 32: 🙈

0xcaff: iirc some of the tests are bindgen struct layout validations, not sure if these were failing or not

samansmink: Ok I will take a look, I think the problem here was that when building with the loadable_extension variant of the bindings, the tests have no way to create a duckdb instance making it basically impossible to run most of the tests

crates/duckdb-loadable-macros/src/lib.rs
```diff
@@ -1,11 +1,84 @@
...
+
+                /// # Safety
+                ///
+                /// Will be called by duckdb
+                #[no_mangle]
+                pub unsafe extern "C" fn #c_entrypoint(info: ffi::duckdb_extension_info, access: ffi::duckdb_extension_access) {
+                    let res = ffi::duckdb_rs_extension_api_init(info, access, #minimum_duckdb_version);
+
+                    if let Err(x) = res {
+                        // Error will be handled by DuckDB
```
0xcaff, line 65: it will?

crates/duckdb/examples/hello-ext-capi/main.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[repr(C)]
+struct HelloBindData {
+    name: *mut c_char,
+}
+
+impl Free for HelloBindData {
+    fn free(&mut self) {
+        unsafe {
+            if self.name.is_null() {
```
0xcaff, line 25: the default value of this pointer is not null, its some undefined value because the memory is allocated in C and passed across the ffi boundary. this branch will probably never be hit. seems like this free trait is implemented everywhere incorrectly in this project though so 🤷  just a heads up

0xcaff: also your code doesn't ever not initialize name

crates/libduckdb-sys/build.rs
```diff
@@ -94,13 +94,21 @@ mod build_bundled {
...
-            let header = HeaderLocation::FromPath(format!("{out_dir}/{lib_name}/src/include/duckdb.h"));
+            let header = HeaderLocation::FromPath(format!("{out_dir}/{lib_name}/src/include/"));
             bindings::write_to_out_dir(header, out_path);
         }
+
         #[cfg(not(feature = "buildtime_bindgen"))]
         {
             use std::fs;
-            fs::copy("src/bindgen_bundled_version.rs", out_path).expect("Could not copy bindings to output directory");
+            fs::copy(
```
0xcaff, line 104: nice can't miss this one

samansmink:
Have a sample extension template here https://github.com/samansmink/extension-template-rs.

I think once we have this PR merged, I can:
- create duckdb/extension-template-rs
- create a new reusable workflow similar to `https://github.com/duckdb/extension-ci-tools/blob/main/.github/workflows/_extension_distribution.yml` but with just latest OS images everywhere since toolchain conpatibility no longer matters.
- Move makefile complexity out of the `extension-template-rs` into `extension-ci-tools` similar to how the c++ extension template does things.
- Add things like a rename script, extension uploading script, etc. 

Then later I need to think about:
- How to setup a testing framework for `extension-template-rs`? 
- Adding this to the community extensions so that people can distribute their Rust-only based extensions.

crates/libduckdb-sys/build.rs
```diff
@@ -265,8 +301,7 @@ mod build_linked {
         // See if pkg-config can do everything for us.
         match pkg_config::Config::new().print_system_libs(false).probe(link_lib) {
             Ok(mut lib) => {
-                if let Some(mut header) = lib.include_paths.pop() {
-                    header.push("duckdb.h");
+                if let Some(header) = lib.include_paths.pop() {
```
0xcaff, line 304: this looks suspicious. noticed i had it in my vendored version but not sure its correct as i don't think i ever exercise the codepath

crates/libduckdb-sys/build.rs
```diff
@@ -305,18 +341,142 @@ mod bindings {
...
+                    None
+                }
+            })
+            .expect("could not find duckdb_ext_api_v0");
+
+        //     let duckdb_ext_api_v0_ident = duckdb_ext_api_v0.ident;
+
+        let p_api = quote::format_ident!("p_api");
+        let mut stores = Vec::new();
+        //     let mut malloc = Vec::new();
```
0xcaff, line 388: might want to get rid of a few of these random commented out lines of code

crates/duckdb-loadable-macros/src/lib.rs
```diff
@@ -1,11 +1,84 @@
...
+            let original_funcname = func.sig.ident.to_string();
+            let prefixed_original_function = func.sig.ident.clone();
+
+            quote_spanned! {func.span()=>
+
+                /// # Safety
+                ///
+                /// Will be called by duckdb
+                #[no_mangle]
+                pub unsafe extern "C" fn #c_entrypoint(info: ffi::duckdb_extension_info, access: ffi::duckdb_extension_access) {
```
0xcaff, line 61: access needs to be a `*const ffi::duckdb_extension_access` instead of a `ffi::duckdb_extension_access`

this somehow works on macos (i think due to the way arguments are passed in ARM on the stack sometimes) but breaks on x86_64 platforms in my testing. painful to debug this one

0xcaff: for reference found it here https://github.com/0xcaff/duckdb_protobuf/pull/17/commits/7467ea6a31050385aabdb1108526dcaad5b9cf99

0xcaff (changes requested):
one blocking correctness issue (the pointer in duckdb_entrypoint_c_api's generated output is missing)

crates/duckdb-loadable-macros/src/lib.rs
0xcaff: i think this is why the windows check is failing currently

0xcaff:
regarding tests for extensions, a contributor contributed tests to my project https://github.com/0xcaff/duckdb_protobuf/blob/master/packages/duckdb_protobuf/tests/it/main.rs might be useful as a scaffold though there are some rough edges to work out still

Makefile
```diff
@@ -12,3 +12,9 @@ all:
 
 test:
 	cargo test --features bundled --features modern-full -- --nocapture
+
+# Build the rust-based CAPI extension demo, change the target arch below to
+EXAMPLE_TARGET_ARCH = osx_arm64
+examples-capi-demo:
+	cargo build --example hello-ext-capi --features="vtab-loadable,loadable_extension" --release
+	python3 scripts/append_extension_metadata.py -l target/release/examples/libhello_ext_capi.dylib -n rusty_quack -dv v0.0.1 -ev v0.0.1 -p $EXAMPLE_TARGET_ARCH
```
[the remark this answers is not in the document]
samansmink: We could do that, but given that we might want to do extension templates for other languages in the future as well, I thought it senses to just have python script to do this instead of a language specific script.

If you feel having a crate for just the extension sigining script makes sense I have no objections for adding them

crates/duckdb-loadable-macros/src/lib.rs
samansmink: Extensions should not throw or exit the program on error, but instead use the method in the extension access struct to set an error, which duckdb will then throw as a C++ exception https://github.com/duckdb/duckdb/blob/2d0842be932e59a887c5decc79ba9d50842be7f7/src/main/extension/extension_load.cpp#L537

samansmink: But now that I think about it I think about it, this is indeed incorrect at the moment:

I should just remove the error potentially returned from the `duckdb_rs_extension_api_init` function:
```rust
if #p_api.is_null() {
    return Err("DuckDB passed a nullpointer while trying to initialize the extension");
}
```

This case does indeed require no extra handling: DuckDB will only return a nullpointer if there is some error. But other parts of the initialization rust side might fail, which requires setting the error using the `extension_access` struct which i'm not doing here yet

crates/duckdb/examples/hello-ext-capi/main.rs
samansmink: I just blindly copied `examples/hello-ext` I will leave it to a future pr to clean this up!

PR #382, opened 2024-09-17
Upgrade arrow-rs to 53

phillipleblanc:
Upgrade arrow-rs to the latest version.

It would be great to have this merged before the next crate version is published for `v1.1.0` (bumped in #381)


Mytherin:
Thanks!

aljazerzen:
This is a breaking change, broke my build.

When bumping major version of arrow, duckdb must bump its major version too.

PR #387, opened 2024-09-29
libduckdb-sys: include duckdb (1.1.1) as a submodule

njaard:
This pull request changes the way libduckdb is built. Because I don't have Windows, I can't verify it works there.

I regenerated the bindings, but the diff was empty. I didn't want libduckdb's jemalloc to conflict with rust's [`jemalloc-sys`
](https://crates.io/crates/jemalloc-sys), so I disabled that extension. It should be documented that you should use the crate [`jemallocator`](https://crates.io/crates/jemallocator) with this crate.

- Build it directly with cmake, having a cargo feature for each extension.
- parquet is documented as being required, so the extension is always enabled and the feature is ignored (for backwards compatibility)
- the "delta" extension doesn't build, so there's no feature for it


breckcs:
I created https://github.com/duckdb/duckdb-rs/issues/461 for further discussion on some needs I have that would benefit from the approach used here.

PR #392, opened 2024-11-07
Minor fixes

samansmink:
Some minor tweaks for the rust extension template:
- extension name and version can now also be passed through env variables: since we need them in the build system for the extension metadata this ensure there's only 1 definition
- also remove an unused import to avoid warnings


Maxxen:
Included as part of #427 (sorry, the squash messed up the history and didnt auto-close this PR)

PR #399, opened 2024-11-13
Upgrade to duckdb 1.1.3

andrewhamon:
Hello! I am hacking on a project using duckdb rust and noticed that the duckdb version was _slightly_ out of date.

This came to my attention because it seems like a community plugin I am trying to use does not have updated binaries for duckdb 1.1.1 (the [bigquery](https://community-extensions.duckdb.org/extensions/bigquery.html) extension, to be specific).

Also, the upgrade script I think was missing `crates/duckdb/Cargo.toml`, so I added that in.


andrewhamon:
I'm a little confused about the intent [of the failing test](https://github.com/duckdb/duckdb-rs/blob/2bd811e7b1b7398c4f461de4de263e629572dc90/crates/duckdb/src/appender/mod.rs#L300-L305).

The panic says "expected error" but I'm wondering if thats a typo and its actually an unexpected error?

andrewhamon:
I've dug into the error a bit more. It seems like in this version of duckdb, the appender no longer errors when writing an incomplete row.

Instead, the append is ignored, and nothing seems to be written to the table (i.e. no rows are returned when querying the table).

If I can get the crate building against a local checkout of duckdb, I'll see if I can bisect when the regression was introduced.

crates/duckdb/src/appender/mod.rs
```diff
@@ -287,22 +289,23 @@ mod test {
...
 
     #[test]
-    fn test_appender_error() -> Result<(), crate::Error> {
+    fn test_appender_computed_columns() -> Result<(), crate::Error> {
         let conn = Connection::open_in_memory()?;
         conn.execute(
             r"CREATE TABLE foo (
             foobar TEXT,
-            foobar_split TEXT[] AS (split(trim(foobar), ','))
+            foobar_twice TEXT AS (foobar || foobar)
```
andrewhamon, line 297: Scalar type makes testing using `query_row` a bit easier (i couldn't figure out how to use query_row for a `(String, Vec<String>)`, got some odd type errors.

Maxxen:
Hello thanks for the PR! Sorry we didn't get to this in time, but now that #427 is merged I don't think this is relevant anymore so Im going to go ahead and close it.

PR #402, opened 2024-11-18
Support building on Android

ndarilek:
* Remove `-stdlib=libstdc++` flag on Android since it breaks the build
* Add `bundled-android-static-libstdcpp` feature for selecting the static C++ runtime (I.e. https://developer.android.com/ndk/guides/cpp-support#use_clang_directly)

I'm open to better name suggestions for `bundled-android-static-libstdcpp`. I don't know enough about clang/LLVM to know what the implications of that flag outside of Android are, but it's specifically mentioned at the above link for selecting a static runtime on Android so I wanted to make the feature clear and explicitly scoped.


ndarilek:
I'd like to merge this if possible. Could I please get approval to run workflows?

Thanks.

mlafeldt:
Thanks for the PR.

Considering that Android support [is still experimental](https://duckdb.org/docs/stable/dev/building/android.html), I'd like to avoid adding special compiler/linker flags that can affect other targets as hard-coded values to build.rs.

The good news is that we use the cc crate in build.rs, which allows you to override specific flags for your particular use case, e.g. via `CXXFLAGS` or `CXXSTDLIB`. See https://docs.rs/cc/latest/cc/#c-support

PR #403, opened 2024-11-20
vtab::arrow: support structs under list and array

ajwerner:


Maxxen:
Hi! Thanks for the PR. Im going to have a closer look at this on Monday and can look at resolving the conflicts then.

ajwerner:
I actually am starting to think that all this code is not the way it should be done. Right now this is effectively duplicating code in rust that duckdb already has. In particular, I became aware a bit later about the Go: `duckdb_arrow_scan` used [here](https://github.com/marcboeker/go-duckdb/blob/686cc28f4e4005ea2c187f16cd832461cea604e8/arrow.go#L269). 

That API is marked as deprecated, but in Discord @Mytherin indicated that it won't just go away without being replaced by something better: https://discord.com/channels/909674491309850675/921073327009853451/1309264012525178952

I can't imagine it makes sense to duplicate this arrow parsing in every library. Same comment goes for https://github.com/duckdb/duckdb-rs/pull/405. 

I feel like a reasonable approach would be to delete all this code and replace it with a call to `duckdb_arrow_scan`, what do you think?

mlafeldt:
https://github.com/duckdb/duckdb/pull/18246 lays the groundwork for a new Arrow C API.

Issue #414, opened 2024-12-19
rfc: Make the Rust vtab interface safe?


sourcefrog:
The VTab trait

https://github.com/duckdb/duckdb-rs/blob/main/crates/duckdb/src/vtab/mod.rs

has

```rust
pub trait VTab: Sized {
    /// The data type of the bind data
    type InitData: Sized + Free;
    /// The data type of the init data
    type BindData: Sized + Free;

    unsafe fn bind(bind: &BindInfo, data: *mut Self::BindData) -> Result<(), Box<dyn std::error::Error>>;
 
    // ....
}
```

This interface is a bit easy to misuse in Rust. The user's implementation of `bind` is called with `data` pointing to an allocated block that is large enough to hold a `Self::BindData` but uninitialized. 

As a result, if you convert the pointer to a `&mut` ref you get undefined behavior. https://github.com/duckdb/duckdb-rs/blob/main/crates/duckdb/examples/hello-ext/main.rs suggests doing this: it _may_ be safe in the specific case of that code, I haven't checked, but if a more realistic implementation has more complex data then it can cause memory corruption.

So, at least, it would be nice to document this more.

But, since this is meant to be the higher-level interface, I think we could do better and just make the interface entirely safe? 

I think there are a couple of potentially feasible approaches. Which one would actually work would take some experimentation:

1.  The Rust `vtab::bind` could just create and return the binddata that it wants, which could then be put in a box and passed to the C API?
2. Or, if `VTab::BindData: Default` then it could be initialized to the default value and passed in as a `&mut`.

I think these would also probably get rid of the need for the `Free` trait there?

This would be an API break but IMHO moving towards a safe interface would be worth it. Although, I suppose, we could probably keep the existing unsafe interfaces present but deprecated.

I can try to send you a PR for this, but I would just like to ask first

1. Is this still maintained (#412)? 
2. Would this be OK in principle?

cc @joewalnes

mlafeldt:
This was already done as part of https://github.com/duckdb/duckdb-rs/pull/428

sourcefrog:
Yep, thanks for merging it!

PR #415, opened 2024-12-21
Safety fixes and docs for vtab

sourcefrog:
The vtab methods are called with pointers to uninitialized blocks of memory:

* The docs previously seemed to falsely indicate that these blocks were actually initialized: clarify the docs.

* It's UB to read uninitialzed memory, and better to write it with ptr::write. In particular, less trivial code that follows the example can corrupt memory if it uses the patterns shown in the example.

* Use unsafe blocks inside unsafe fns in the example, because this will be a warning in Rust 2024. In particular, do this to avoid a warning from the duckdb_entrypoint macro.

* The Rust types used in the examples don't need layout compatibility with C, and can simply be Rust structs using native strings.

Improves #414


Maxxen:
Thanks a lot of the PR, I've Included it as part of #427 (sorry, the squash messed up the history and didnt auto-close this PR)

PR #416, opened 2024-12-21
Make the VTab interface safe

sourcefrog:
With this PR, simple table function extensions no longer need any unsafe code, and so are easier to implement and less likely to crash the process.

This partially solves #414.

I have not yet converted every interface to be safe, but if you like the general approach then I could go on to update everything in `crates/duckdb/src/vtab/function.rs`.

This is an API break for Rust extensions. In my opinion it's worth it because the existing API is easy to misuse and it would be better for callers to use a safe interface. It would not be impossible to preserve the existing unsafe interface, but I think it would leave a fair amount of complexity both in the implementation and in the user-facing API. 

Rather than passing a pointer to a block of uninitialized memory, which can easily lead to UB, the setup functions now just return Rust objects.

The `Free` trait is no longer required, as the Rust type will now just be dropped. If extensions want custom behavior on drop, such as closing a connection, they can do that from `impl Drop`.

It looks like the core DuckDB engine will, in at least some cases, run the table function from multiple threads, so to make that safe I've made the references always shareable. Extensions can use mutexes or atomic types internally as desired.

This builds on and goes further than #415.

### DIFF
--- a/crates/duckdb-loadable-macros/src/lib.rs
+++ b/crates/duckdb-loadable-macros/src/lib.rs
@@ -133,8 +133,10 @@ pub fn duckdb_entrypoint(_attr: TokenStream, item: TokenStream) -> TokenStream {
                 /// Will be called by duckdb
                 #[no_mangle]
                 pub unsafe extern "C" fn #c_entrypoint(db: *mut c_void) {
-                    let connection = Connection::open_from_raw(db.cast()).expect("can't open db connection");
-                    #prefixed_original_function(connection).expect("init failed");
+                    unsafe {
+                        let connection = Connection::open_from_raw(db.cast()).expect("can't open db connection");
+                        #prefixed_original_function(connection).expect("init failed");
+                    }
                 }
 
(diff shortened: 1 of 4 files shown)


Maxxen:
Cool, I think this make sense. Although I imagine we could pass the init state to the callback already wrapped in an RwLock just to be extra safe instead of having the user deal with synchronization. But I think this is fine for now as it is already an improvement.

sourcefrog:
> Cool, I think this make sense. Although I imagine we could pass the init state to the callback already wrapped in an RwLock just to be extra safe instead of having the user deal with synchronization. But I think this is fine for now as it is already an improvement.

Yep, there is more that we could do; I think ideally the whole interface would be safe including with regard to multithreading from C++. I just didn't want to sink too much time before confirming that it could in principal be merged, and in particular that dropping the `unsafe` interfaces was OK.

Maxxen:
Merged as part of https://github.com/duckdb/duckdb-rs/pull/428!

PR #424, opened 2025-01-31
Fix test  to trigger append error

bnmoch3:
This PR fixes the `test_appender_error` test, which was failing due to changes in DuckDB’s handling of virtual generated columns. DuckDB treats `foobar_split` as a virtual generated column, meaning its value is computed at query time and cannot be inserted into. Previously, the test assumed appending to the `foobar` column alone would trigger an error, but DuckDB allowed the insertion, causing the test to fail. By adding `foobar_int`, the test now correctly validates that appending an incomplete row still fails as expected.

### DIFF
--- a/crates/duckdb/src/appender/mod.rs
+++ b/crates/duckdb/src/appender/mod.rs
@@ -288,20 +288,23 @@ mod test {
 
     #[test]
     fn test_appender_error() -> Result<(), crate::Error> {
+        use crate::params;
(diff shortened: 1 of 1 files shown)


Maxxen:
Thanks a lot of the PR, I've Included it as part of #427 (sorry, the squash messed up the history and didnt auto-close this PR)

bnmoch3:
You're welcome :)

Issue #436, opened 2025-02-14
Upgrading `duckdb-rs` to 1.2.0 fails linking on Ubuntu 22.04


cpaika:
When trying to upgrade duckdb-rs from 1.1.1 to 1.2.0, we're getting compilation failures on our Linux build machines when using the `bundled` flag.  It looks like duckdb compiles successfully, but linking fails into the final binary.  This build is working with duckdb 1.1.1, a change in 1.2.0 is somehow breaking it.

```
 > [8/8] RUN $HOME/.cargo/bin/cargo test:
266.3   = note: /usr/bin/ld: /app/target/debug/deps/liblibduckdb_sys-f5c8fd0dd29e7ad5.rlib(d72e143842620a4d-ub_src_planner_expression.o): in function `void duckdb::DynamicCastCheck<duckdb::ListLambdaBindData, duckdb::FunctionData>(duckdb::FunctionData const*)':
266.3           /app/target/debug/build/libduckdb-sys-51da510693e3ab30/out/duckdb/src/include/duckdb/common/helper.hpp:262: undefined reference to `typeinfo for duckdb::ListLambdaBindData'
266.3           /usr/bin/ld: /app/target/debug/build/libduckdb-sys-51da510693e3ab30/out/duckdb/src/include/duckdb/common/helper.hpp:262: undefined reference to `typeinfo for duckdb::ListLambdaBindData'
266.3           collect2: error: ld returned 1 exit status
266.3
266.3   = note: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
266.3   = note: use the `-l` flag to specify native libraries to link
266.3   = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-lib)
266.3
266.4 error: could not compile `duckdb-repro` (bin "duckdb-repro" test) due to 1 previous error
Dockerfile:23
```

I've included a minimal reproduction of the issue here: https://github.com/cpaika/duckdb-reproduction/tree/main

My guess is this is related to RTTI, or the LambdaBindGen object isn't being included in the final artifact somehow but not sure.

Any ideas what the problem?  Or if there's a compilation flag to fix it?

jessekrubin:
ditto! same thing here!

Maxxen:
[...]

PR #438, opened 2025-02-17
fix: pin duckdb and libduckdb-sys versions

luong-komorebi:
**Context** :
- Rust considers a package version to be compatible if their leftmost greater-than-zero version is the same. This means 1.2.0 is compatible with 1.1.1. This is not "wrong" and complies well with "semver". However, minor releases of duckdb usually contains breaking changes  (see changelogs at https://r.duckdb.org/news/index.html) and 1.2.0 might break 1.1.1
- Recent changes to libduckdb-sys breaks build process on ubuntu for duckdb-rs https://github.com/duckdb/duckdb-rs/issues/436. Since there was no version lock for libduckdb-sys, there are cases that even when locking duckdb-rs in downstream repos, libduckdb-sys is still upgraded to the newer and mismatched version. 

**Proposed changes**:
- Strictly lock version of libduckdbsys and duckdb by using [Comparision Requirement](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#comparison-requirements). While tilde requirement can be used as well, I am not too sure if that is suitable given that every version bump in this repo usually couples duckdb-rs and libduckdb-sys

**Note**:
- This should be backported to version `1.1.1` as `1.1.2` to safeguard maintainers from accidentally being bumped to 1.2.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,8 @@ license = "MIT"
 categories = ["database"]
 
 [workspace.dependencies]
-duckdb = { version = "1.2.0", path = "crates/duckdb" }
-libduckdb-sys = { version = "1.2.0", path = "crates/libduckdb-sys" }
+duckdb = { version = "=1.2.0", path = "crates/duckdb" }
+libduckdb-sys = { version = "=1.2.0", path = "crates/libduckdb-sys" }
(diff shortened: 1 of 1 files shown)


Maxxen:
Hello! Thanks a lot for the PR! I've included these changes as part of https://github.com/duckdb/duckdb-rs/pull/462, so I'm going to go ahead and close this.

PR #439, opened 2025-02-18
Add support for Map writing to Arrow vtab

sgrebnov:
PR adds support for writing Map Arrow data type to Arrow vtab.

`MapArray` is physically implemented as a `ListArray` of structs with key and value fields, and support for this can be added using `list_vector` and `struct_array_to_vector`.

While working on large Map arrays, I discovered that `StructVector.child` must adjust its vector capacity similarly to `ListVector.child`; otherwise it can reach array capacity (`"assertion failed: array.len() <= out.capacity()"`). Improved.

~Note: the lint error is unrelated to this change and due to recently released `rust version 1.85 / 2025-02-17` that automatically applied to the CI (requires improvement in `crates/duckdb/src/r2d2.rs:118:31`)~
 - https://github.com/duckdb/duckdb-rs/pull/443

### DIFF
--- a/crates/duckdb/src/core/vector.rs
+++ b/crates/duckdb/src/core/vector.rs
@@ -192,7 +192,8 @@ impl ListVector {
     }
 
     /// Take the child as [StructVector].
-    pub fn struct_child(&self) -> StructVector {
+    pub fn struct_child(&self, capacity: usize) -> StructVector {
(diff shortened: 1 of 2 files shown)


Maxxen:
Thanks a lot for the PR! LGTM

PR #440, opened 2025-02-22
Deploy the API doc to GitHub Pages

yutannihilation:
Usual APIs can be found on the documents on docs.rs. However, optional items like `VScalar` and `VTab` are not found, while we need such things to write an extension. Does it make sense to build the doc with the necessary feature flags and deploy it to GitHub Pages?

Example:
https://yutannihilation.github.io/duckdb-rs/?search=VTab

### DIFF
--- a/.github/workflows/extension-docs.yml
+++ b/.github/workflows/extension-docs.yml
@@ -44,16 +44,13 @@ jobs:
           # cf. https://github.com/actions/deploy-pages/issues/188#issuecomment-1597651901
           rm -f ./target/doc/.lock
 
-          mkdir _docs
-          mv ./target/doc ./_docs/extension-api
-
       - name: Setup Pages
         uses: actions/configure-pages@v4
 
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v3
         with:
-          path: './_docs/'
+          path: './target/doc'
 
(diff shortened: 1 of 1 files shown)


yutannihilation:
Thanks, it might be an option. But, I'm afraid the APIs with and without `loadable-extension` might be different because it serves different purposes; C API (controlling DuckDB itself) vs C extension API (implementing an extension). I think docs.rs should primarily show the one corresponding to C API. That said, I don't know well about the differences, so I might be wrong here.

https://github.com/duckdb/duckdb-rs/blob/6d5368011043bd65dac42cf5c7a26784c16d82cc/crates/libduckdb-sys/build.rs#L182-L214

Maxxen:
Hello! Thanks a lot for the PR! I think this is a general good idea but I need to investigate a bit more on how we want to handle docs so I'll get back to this later.

yutannihilation:
Thanks for the update! Please feel free to close this if you find some better way.

yutannihilation:
> optional items like `VScalar` and `VTab` are not found,

Now v1.2.1 document on docs.rs contains these items, so my problem is resolved, thanks!

I believe it's still useful to use GitHub pages to serve the dev version of the document. But, let's revisit here in future. I'm closing this pull request for now.

PR #447, opened 2025-02-27
A temporary fix for compile error with chrono v0.4.40

yutannihilation:
Due to an unfortunate collision, arrow-rs currently doesn't compile without pinning the version of chrono to before v0.4.40. It's already fixed in the dev version of arrow-rs, but it takes some time to be released on crates.io. So, this pull request adds a temporary fix by specifying the version. After the new version of arrow-rs is released, this tweak can be removed (but probably the versions should be bumped to the latest to make sure the appropriate versions are chosen).

c.f. https://github.com/apache/arrow-rs/issues/7196

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ byteorder = "1.3"
 calamine = "0.22.0"
 cast = "0.3"
 cc = "1.0"
-chrono = "0.4.22"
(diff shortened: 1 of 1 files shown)


yutannihilation:
Closing in favor of https://github.com/duckdb/duckdb-rs/pull/448

PR #448, opened 2025-02-27
Bump arrow to the version compatible with chrono 0.4.40

yutannihilation:
Due to an unfortunate collision with the latest chrono crate, the arrow crate before version 54.2.1 doesn't compile. Although the version 54.2.1, which addressed the compatibility issue, is about to be released (~~note: at the time of writing this, it's not released yet~~  Released!!!), it's probably a good idea to bump the versions of chrono and arrow to make sure the compilation error won't happen.

c.f.)
- problem: https://github.com/apache/arrow-rs/issues/7196
- the tracking issue of arrow's new release: https://github.com/apache/arrow-rs/issues/7209


alamb:
FYI we have released a fix in 54.2.1: https://github.com/apache/arrow-rs/issues/7209

Cargo.toml
```diff
@@ -28,7 +28,7 @@ byteorder = "1.3"
 calamine = "0.22.0"
 cast = "0.3"
 cc = "1.0"
-chrono = "0.4.22"
+chrono = "0.4.40"
```
alamb, line 31: the hotfix for arrow basically encodes this requirement in arrows cargo.toml

PR #454, opened 2025-03-04
Fix #380, bug in code that reads Enums 

### LINKED ISSUE #380: Bug in code that reads Enums
It's a coincidence that the current enum tests in `test_all_types.rs` pass.   Here's a simple example that shows a bug due to incorrect array indexing.

```rust
use duckdb::{Connection, Result};

#[derive(Debug)]
struct Person { id: i32, name: String, }

fn main() -> Result<()> {
    let conn = Connection::open_in_memory()?;
    conn.execute_batch(r#"
CREATE TABLE enum_test (
    id INTEGER,
    name ENUM('AA', 'BB', 'CC')
);
INSERT INTO enum_test VALUES (0, 'BB');
INSERT INTO enum_test VALUES (1, 'AA');
INSERT INTO enum_test VALUES (2, 'CC');
INSERT INTO enum_test VALUES (3, 'BB');
    "#)?;
    let mut stmt = conn.prepare("SELECT id, name FROM enum_test")?;
    let person_iter = stmt.query_map([], |row| {
        Ok(Person{
            id: row.get(0)?,
            name: match row.get_ref_unwrap(1).to_owned() {
                duckdb::types::Value::Enum(e) => e,
                _ => panic!("Oops")
            },
        })
    })?;

    for person in person_iter {
        println!("Found person {:?}", person.unwrap());
    }
    Ok(())
}
```
Here is the error
```
Found person Person { id: 0, name: "AA" }
Found person Person { id: 1, name: "BB" }
Found person Person { id: 2, name: "CC" }
thread 'main' panicked at /home/adrian/.cargo/registry/src/index.crates.io-6f17d22bba15001f/arrow-array-52.1.0/src/array/byte_array.rs:305:9:
Trying to access an element at index 3 from a StringArray of length 3
stack backtrace: [...]
```
It looks like the indexing loops over the enum values instead of the actual row values.   This is related to the #365 issue I reported on 29Jul.  I am quite sure now it's a bug in the Rust DuckDb driver. 

Best regards,
Tony

fotonick:
In turning a `ValueRef::Enum` to a `Value`, the code was inappropriately indexing with the row index directly into the dictionary's string array, so iterating any query's results was effectively just listing the Enum variants in order until row index > number of variants and it panicked. Fix by getting the row/col value (the dictionary key) and using that to index into the dictionary's string array.

You can see it in all three branches of the match statement at [value_ref.rs#L245](https://github.com/duckdb/duckdb-rs/blob/1096461ea9d6a9632e5cb079860ea47b2de0257b/crates/duckdb/src/types/value_ref.rs#L245). These call [DictionaryArray.values()](https://github.com/apache/arrow-rs/blob/3674073bba31e082b337cee63726cfeb186a3a0c/arrow-array/src/array/dictionary_array.rs#L366), whose docstring says clearly that it returns the dictionary values.

Note that this code was covered by a test, but the expected value was hardcoded incorrectly to make the previous implementation pass. Proof that the old expected value was incorrect and the new one is correct:
```sql
D SELECT medium_enum FROM test_all_types();
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                 medium_enum                                                                  │
│ enum('enum_0', 'enum_1', 'enum_2', 'enum_3', 'enum_4', 'enum_5', 'enum_6', 'enum_7', 'enum_8', 'enum_9', 'enum_10', 'enum_11', 'enum_12', …  │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ enum_0                                                                                                                                       │
│ enum_299                                                                                                                                     │
│ NULL                                                                                                                                         │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

### DIFF
--- a/crates/duckdb/src/test_all_types.rs
+++ b/crates/duckdb/src/test_all_types.rs
@@ -651,7 +651,7 @@ fn test_single(idx: &mut i32, column: String, value: ValueRef) {
         },
         "medium_enum" => match idx {
             0 => assert_eq!(value.to_owned(), Value::Enum("enum_0".to_string())),
-            1 => assert_eq!(value.to_owned(), Value::Enum("enum_1".to_string())),
+            1 => assert_eq!(value.to_owned(), Value::Enum("enum_299".to_string())),
             _ => assert_eq!(value, ValueRef::Null),
         },
         "large_enum" => match idx {
--- a/crates/duckdb/src/types/value_ref.rs
+++ b/crates/duckdb/src/types/value_ref.rs
@@ -6,8 +6,8 @@ use rust_decimal::prelude::*;
 
 use arrow::{
     array::{
-        Array, ArrayRef, DictionaryArray, FixedSizeListArray, LargeListArray, ListArray, MapArray, StructArray,
-        UnionArray,
+        Array, ArrayRef, DictionaryArray, FixedSizeListArray, LargeListArray, ListArray, MapArray, StringArray,
+        StructArray, UnionArray,
     },
     datatypes::{UInt16Type, UInt32Type, UInt8Type},
 };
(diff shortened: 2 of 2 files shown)


fotonick:
I should mention that `cargo test --features bundled` passes, as does `ASAN_OPTIONS=detect_leaks=1 ASAN_SYMBOLIZER_PATH=/opt/homebrew/Cellar/llvm/19.1.7/bin/llvm-symbolizer cargo test --features bundled -- --nocapture`.

Maxxen:
Hello! Thanks a lot for the PR! I've included these changes as part of https://github.com/duckdb/duckdb-rs/pull/462, so I'm going to go ahead and close this.

PR #456, opened 2025-03-06
Fix writing Arrow Decimal128 with zero scale

sgrebnov:
DuckDB does not support negative scale, but the existing logic required the scale to be positive and incorrectly handled Decimal128 with zero scale. This PR fixes support for zero scale and adds an explicit error in case a Decimal with a negative scale is specified.

Added unit test for Decimal128 with zero scale.


sgrebnov:
@Maxxen - if there is `duckdb-rs` 1.2.1 release, can this be included in that release please. Thank you!

Maxxen:
Hello! Thanks for the PR! I've included it as part of https://github.com/duckdb/duckdb-rs/pull/462

PR #488, opened 2025-04-09
Implement support for `duckdb_arrow_scan` for ingesting data via Arrow

phillipleblanc:
Adds support for calling the duckdb_arrow_scan when passed an FFI_ArrowArrayStream from the Arrow library that implements the https://arrow.apache.org/docs/format/CStreamInterface.html

This matches behavior that exists in the Go DuckDB library: https://github.com/marcboeker/go-duckdb/blob/c607539e645091c1e3f600f929229d8baa7166e4/arrow.go#L251

This also provides an alternative to the [`arrow-vtab`](https://github.com/duckdb/duckdb-rs/blob/main/crates/duckdb/src/vtab/arrow.rs#L103) module that is not zero-copy in all cases for Arrow -> DuckDB buffers, whereas the native C++ DuckDB code has better support for zero-copy Arrow -> DuckDB buffers.

### DIFF
--- a/crates/duckdb/src/arrow_scan.rs
+++ b/crates/duckdb/src/arrow_scan.rs
@@ -1,0 +1,233 @@
+use std::ffi::CString;
+
+use arrow::ffi_stream::FFI_ArrowArrayStream;
+
+use crate::{error::error_from_duckdb_code, ffi, Connection, Error, Result};
+
+impl Connection {
+    /// Registers a temporary view in DuckDB based on an Arrow stream.
+    ///
+    /// Note the underlying `duckdb_arrow_scan` C API is marked for deprecation.
+    /// However, similar functionality will be preserved in a new yet-to-be-determined API.
+    ///
+    /// # Arguments
+    ///
+    /// * `view_name`: The name of the view to register
+    /// * `arrow_scan`: The Arrow stream to register
+    pub fn register_arrow_scan_view(&self, view_name: &str, arrow_scan: &FFI_ArrowArrayStream) -> Result<()> {
+        let conn = self.db.borrow_mut().con;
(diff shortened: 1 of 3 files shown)


Maxxen:
Hello! Thanks a lot for the PR! Im about to cut v1.2.2 so in order to expedite getting this merged I forked this off into #489 with formatting/clippy fixes

Maxxen:
Actually, ASAN reports memory leaks related to the test. I've done some digging and asked around the office from people with more arrow-knowledge, and my impression is that the `arrow_scan_array` method is most likely partly broken in that it has some weird assumptions regarding ownership and doesn't always free schemas properly. This PR [seems related](https://github.com/duckdb/duckdb/pull/16050).

I guess these functions are deprecated for a reason. I don't know exactly when we will get around to replace them with something better, but we're looking into trying to fix what we think is causing _this_ issue in particular - and if we do I'll check against this PR again.

phillipleblanc:
> Actually, ASAN reports memory leaks related to the test. I've done some digging and asked around the office from people with more arrow-knowledge, and my impression is that the `arrow_scan_array` method is most likely partly broken duckdb-side in that it has some weird assumptions regarding ownership and doesn't always free schemas properly. This PR [seems related](https://github.com/duckdb/duckdb/pull/16050).
> 
> I guess these functions are deprecated for a reason. I don't know exactly when we will get around to replace them with something better, but we're looking into trying to fix what we think is causing _this_ issue in particular - and if we do I'll check against this PR again.

Thanks for checking! I'll convert this PR to draft until the memory leak issue is fixed and/or a new C API is available that we can use.

phillipleblanc:
Looks like the new C API will be coming out in DuckDB v1.4. I'll close this PR since we'll need a new one to integrate with the new C API.

PR #493, opened 2025-04-11
Add interrupt handle

jankramer:
This PR adds an `InterruptHandle` that allows interrupting long-running queries from another thread. Internally, this calls the `duckdb_interrupt` function from the C API. The implementation is inspired by [rusqlite][1].

It seems to get the job done, but a couple of thoughts:

1. Testing the interrupts is a bit tricky. I wasn't able to come up with a deterministic approach, so I took inspiration from tests in other client APIs and simply kick off an expensive query.
2. Unfortunately, the error code returned on interrupt is `ErrorCode::Unknown`, so we have to resort to a match on the error message. Works for now, but worth looking into perhaps.
3. I'm not super confident in the unsafe bits, so a critical pair of eyes is welcome.

Related to #343, but that PR got stuck.

[1]: https://docs.rs/rusqlite/latest/rusqlite/struct.InterruptHandle.html


mlafeldt (approved):

mlafeldt:
@jankramer Thanks a bunch. This looks great!

Regarding the error code, libduckdb-sys currently always sets it to `Unknown`.

```rust
impl Error {
    pub fn new(result_code: duckdb_state) -> Self {
        Self {
            code: ErrorCode::Unknown,
            extended_code: result_code,
        }
    }
}
```

I don't know why that is, but it's definitely an area that could be improved in the future.

PR #504, opened 2025-05-22
Mark no_mangle attribute as unsafe

gatesn:
As per https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-attributes.html


mlafeldt:
This commit was merged as part of #505 and shipped in DuckDB 1.3.0.

PR #505, opened 2025-05-23
[WIP] Update to DuckDB v1.3.0

Maxxen:


Maxxen:
This will fail due to changes in the arrow decimal conversion code in DuckDB. Will have to consult with the arrow-knowledgeable folks at labs.

wyhaya:
@Maxxen Hi, can add this in libduckdb-sys/build.rs? This can fix the build error on Windows.

```rust
println!("cargo:rustc-link-lib=dylib=Rstrtmgr");
println!("cargo:rustc-link-lib=dylib=Bcrypt");
```

PR #511, opened 2025-06-11
feat(types): Add into_inner and into_inner_as methods for Value type

izumkineno:
I can not find methods to get my data from Value, so I write this, but may be need test

```
    let rows = stmt
        .query_map([], |row| {
            use duckvfslink::duckdb::types::Value;

            let id: String = row.get(0).unwrap();
            let pid: String = row.get(1).unwrap();
            let fid: String = row.get(2).unwrap();
            let ps: Value = row.get(3).unwrap();
            let is_active: bool = row.get(4).unwrap();
            let is_dir: bool = row.get(5).unwrap();

            let t = ps.into_inner_as::<Vec<Value>>().unwrap();
            let ps: Vec<String> = t
                .into_iter()
                .map(|v| v.into_inner_as::<String>())
                .filter_map(|v| v)
                .collect::<Vec<String>>();

            Ok(TreeViewNode {
                id,
                pid,
                fid,
                ps,
                is_active,
                is_dir,
            })
        })
        .unwrap()
        .filter_map(|r| r.ok())
        .collect::<Vec<_>>();
```


mlafeldt (changes requested):
Thanks!

Opposed to the title, this PR does not add `into_inner`. Instead, it adds `into_inner_as` and serialization support. Please move the latter to another PR. I'd love to know your actual use case!

In both cases, some tests would be appreciated. 🙏

PR #530, opened 2025-07-08
Support appending large Arrow record batches

mlafeldt:
Based on the test from #271 by @plaflamme, this PR enhances/fixes `append_record_batch` to automatically chunk the passed record batch into smaller pieces that fit within DuckDB's vector size limit (using Arrow's zero-copy slicing).


Maxxen (approved):
Great work!

PR #558, opened 2025-07-31
Support passing custom state to scalar functions

### LINKED ISSUE #506: is there any way to add extra_info to scalar funtion or table funtion.
I need to register glabal state to scalar function or table function.
but I found that extra_info for register operation is not easy to understand.

Can anyone provide some examples to use extra_info in extension to share state or struct.

thanks.

mlafeldt:
The current API isn't very helpful for passing state to scalar functions, as it's limited to `State::default()`. This PR adds support for arbitrary data.

Introduces a breaking change to the low-level `set_extra_info` helper, which is acceptable IMO.

Fixes #506

### DIFF
--- a/crates/duckdb/src/vscalar/mod.rs
+++ b/crates/duckdb/src/vscalar/mod.rs
@@ -25,7 +25,7 @@ pub use arrow::{ArrowFunctionSignature, ArrowScalarParams, VArrowScalar};
 pub trait VScalar: Sized {
     /// State that persists across invocations of the scalar function (the lifetime of the connection)
     /// The state can be accessed by multiple threads, so it must be `Send + Sync`.
-    type State: Default + Sized + Send + Sync;
+    type State: Sized + Send + Sync;
     /// The actual function
     ///
     /// # Safety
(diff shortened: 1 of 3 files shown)


crates/duckdb/src/vscalar/mod.rs
```diff
@@ -132,15 +132,32 @@ where
...
-            scalar_function.set_extra_info::<S::State>();
+            scalar_function.set_extra_info(S::State::default());
+            set.add_function(scalar_function)?;
+        }
+        self.db.borrow_mut().register_scalar_function_set(set)
+    }
+
+    /// Register the given ScalarFunction with custom state
+    #[inline]
+    pub fn register_scalar_function_with_state<S: VScalar>(&self, name: &str, state: S::State) -> crate::Result<()>
```
Maxxen, line 151: I might be too c++-brained, but if we're going to clone the state into the box anyway, why not pass a reference here?

mlafeldt: You're absolutely right! Done.

Maxxen (approved):
LGTM!

Issue #591, opened 2025-09-17
`now() - interval '1' day` returns an error


krinart:
Following code will return error:
```rust
use duckdb::{params, Connection, Result};

fn main() -> Result<()> {
    let conn = Connection::open_in_memory()?;
    conn.prepare("select now() - interval '1 day'")?;
    Ok(())
}
```

Error:
```
Binder Error: No function matches the given name and argument types '-(TIMESTAMP WITH TIME ZONE, INTERVAL)'. You might need to add explicit type casts.
Candidate functions:
...
-(DATE, DATE) -> BIGINT
-(DATE, INTEGER) -> DATE
-(TIMESTAMP, TIMESTAMP) -> INTERVAL
-(INTERVAL, INTERVAL) -> INTERVAL
-(DATE, INTERVAL) -> TIMESTAMP
-(TIME, INTERVAL) -> TIME
-(TIMESTAMP, INTERVAL) -> TIMESTAMP
-(TIME WITH TIME ZONE, INTERVAL) -> TIME WITH TIME ZONE
-(INTERVAL) -> INTERVAL
```

As you can see there's indeed no matching function.

However the same query works through duckdb cli:

<img width="328" height="116" alt="Image" src="https://github.com/user-attachments/assets/a2cdd79d-79af-4c54-9228-6dc784d50148" />

Versions:
* cli: 1.3.2
* duckdb-rs: 1.3.2, 1.4.0

mlafeldt:
The DuckDB CLI has ICU statically linked and loaded by default, while the bundled version in duckdb-rs doesn't automatically load ICU. The workaround is to load ICU manually. Please see https://github.com/duckdb/duckdb-rs/issues/500#issuecomment-3134271500

sgrebnov:
Should match DuckDB CLI behavior where `ICU ` is STATICALLY_LINKED?

Similar to `core_functions` that was recently bundled
https://github.com/duckdb/duckdb-rs/commit/5eeb1f01c278790ce1e2d24045f0096e9e2528e4

PR #594, opened 2025-09-19
Add ICU extension to the list of statically linked extensions

krinart:
## 🗣 Description

Query `select now() - interval '1'` works via duckdb CLI, but fails via `duckdb-rs` without manually installing/loading ICU extension. In order to match the behavior of CLI, we can make it statically linked.

## 🔨 Related Issues

https://github.com/duckdb/duckdb-rs/issues/591


mach-kernel (approved):

Maxxen:
I dont think this will work, Ive attempted to do this before, but it brings the total source code size to larger than what crates.io allows, making the package upload fail.

PR #598, opened 2025-10-03
Support multi-statement queries

### LINKED ISSUE #310: Pivot not working with prepared statements
Executing the pivot statement results in the following error:
```rust
let sql = r"
with tbl as (
  select * from read_parquet('test.parquet')
)
PIVOT tbl on a using sum(b)";
let con = duckdb::Connection::open_in_memory().unwrap();
let stmt = con.prepare(sql);
println!("{:?}", stmt);
// Err(DuckDBFailure(Error { code: Unknown, extended_code: 1 }, Some("Invalid Input Error: Cannot prepare multiple statements at once!")))
```
It may be related to https://github.com/duckdb/duckdb/issues/7720

### LINKED ISSUE #425: Pivot not working
Below query is not working but it is working in [shell.duckdb.com](https://shell.duckdb.org/)
Doc at https://duckdb.org/docs/sql/statements/pivot.html

```sql
PIVOT cities
ON year
USING sum(population);
```

Why it is not working with this library? I am getting below error.

```
invalid Input Error: Cannot prepare multiple statements at once!
```

mlafeldt:
Inspired by the [Go client](https://github.com/mlafeldt/go-duckdb/blob/2227e78d82a5e48b320f412cae5e572a0620d98e/connection.go#L203-L235), this enhances `prepare` to handle both single and multi-statement queries.

Among other things, this enables the use of `PIVOT` (which internally expands to multiple statements) and also enables multiple statements per line in the example REPL. 🎉

Fixes #425
Fixes #310


Maxxen (approved):
LGTM, great work!

PR #604, opened 2025-10-20
Add jiff support

StratusFearMe21:
This PR adds initial `jiff` support to these bindings. It's a newer date/time library inspired by JavaScript's Temporal API. `jiff` is unique in that it has a `Span` type that can accurately perform arithmetic with calendar units like days and moths. Jiff's span type also losslessly converts to and from DuckDB's interval type.


StratusFearMe21:
This PR also has some fixes for the implementation of `chrono`. For instance, the `FromSql` implementation for `chrono::NaiveTime` and `chrono::NaiveDate` used `chrono::NaiveDateTime::parse_from_string()`, which crashes if the underlying `ValueRef` is a string, since you need both a date and a time to construct a `NaiveDateTime`. I've made sure to add more unit tests for the bugs I did find.

PR #607, opened 2025-10-23
Add support for DEFAULT values in Appender

### LINKED ISSUE #551: Appender: support appending default column values
Over the past few days, I've been using duckdb-rs, and for sure, there are indeed still some issues needed figured out. 

When using the Appender to handle default values, it doesn't work well.

At least during my recent testing, trying to insert `None::<T>, duckdb::types::Value::Null, or leaving the value unspecified` all led to various errors—often resulting in  `fatal runtime error: Rust cannot catch foreign exceptions, aborting`.

I haven’t seen a similar issue reported in duckdb-rs, but there are many related issues in duckdb. Fortunately, this particular issue has already figured out on June 7, 2024. 

So maybe we need a `AppendDefault`.

See [PR#11905](https://github.com/duckdb/duckdb/pull/11905), [discussions#9158](https://github.com/duckdb/duckdb/discussions/9158), [issues#14343](https://github.com/duckdb/duckdb/issues/14343) for more.

mlafeldt:
Implements `AppendDefault` marker type that allows appending a column's `DEFAULT` value (as defined in table schema) when using the Appender API.

Fixes #551

### DIFF
--- a/crates/duckdb/src/types/to_sql.rs
+++ b/crates/duckdb/src/types/to_sql.rs
@@ -8,6 +8,12 @@ use std::borrow::Cow;
 /// default values (as defined in the table schema). Unlike `Null` which explicitly
 /// sets a column to NULL, `AppendDefault` uses the column's DEFAULT expression.
 ///
+/// ## Limitations
+///
+/// `AppendDefault` only works with **constant** default values. Non-deterministic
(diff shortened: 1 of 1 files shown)


mlafeldt:
Back to draft status. Looking into `duckdb_append_default_to_chunk` as an alternative.

PR #615, opened 2025-10-29
Implement Params for HashMap and serde_json::Value

ryanschneider:
Thanks for implementing #609, that gave me enough to take a stab at implementing `Params` for `serde_json::Value` and `HashMap`.  Let me know what you think!

### DIFF
--- a/crates/duckdb/src/params.rs
+++ b/crates/duckdb/src/params.rs
@@ -1,4 +1,5 @@
-use crate::{Result, Statement, ToSql};
+use crate::{Error, Result, Statement, ToSql};
+use std::collections::HashMap;
 
 mod sealed {
     /// This trait exists just to ensure that the only impls of `trait Params`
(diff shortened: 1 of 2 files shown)


crates/duckdb/src/params.rs
```diff
@@ -303,3 +334,65 @@ where
...
+    impl Params for &Value {
+        fn __bind_in(self, stmt: &mut Statement<'_>) -> crate::Result<()> {
+            match self {
+                Value::Object(ref map) => {
+                    let n = stmt.parameter_count();
+                    let params: Vec<&dyn ToSql> = (1..=n)
+                        .map(|i| {
+                            let name = stmt.parameter_name(i)?;
+                            let val = map.get(&name).ok_or(Error::InvalidParameterName(name))?;
+                            Ok(val as &dyn ToSql)
```
mlafeldt: I'm afraid this will serialize everything as strings because of `serde_json::to_string(self).unwrap()`.

mlafeldt (changes requested):
The HashMap impl is solid. The serde_json impl has a fundamental type-safety problem: ToSql for serde_json::Value serializes everything to JSON strings. A fix would require Value variants to be converted to native Rust types before binding.

I suggest shipping the HashMap now and defer (delete) serde_json support since it's a convenience on top.

mlafeldt:
I just merged #748, which brings your HashMap support + a nice new `named_params!` macro to duckdb-rs. 🚀

ryanschneider:
Thanks I got pulled into a different project so didn't have time to follow up but thanks for making the changes and getting it merged!

PR #623, opened 2025-11-07
Create an appender to specific catalog + schema + table

### LINKED ISSUE #622: Issue with Appender
Hello, on Duckdb Connection struct, there is a method `appender_to_db` that returns an Appender for the specified table in the provided schema.

However, it will not work if there is a catalog attached with a same name as the schema.
And there is no way to provide the catalog name as parameter for creating the Appender.

I tried passing it alongside the schema like this: catalog_name.schema with no success. I tried quoting both catalog_name and schema, quoting the whole string. No success.

Maybe the issue is with core duckdb project but I’m not very literate in C++.

By the way, I’m new to Duckdb and Duckdb-rs. Thank you for this very cool project!

Thanks for you help.

mchataigner:
Leverage duckdb core duckdb_appender_create_ext function that does exactly that. This avoids errors due to ambiguous names between an attached catalog and a schema.

Fixes #622

### DIFF
--- a/crates/duckdb/src/inner_connection.rs
+++ b/crates/duckdb/src/inner_connection.rs
@@ -176,6 +176,31 @@ impl InnerConnection {
         Ok(Appender::new(conn, c_app))
     }
 
+    pub fn appender_to_catalog_and_db<'a>(
+        &mut self,
+        conn: &'a Connection,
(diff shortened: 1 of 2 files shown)


Maxxen:
Thanks!

PR #627, opened 2025-11-15
fix: expose is_invalid to avoid panic if the logical type is invalid

rubenfiszel:
We got quite burned by this for queries where id() would just panic.
Example of a query that would panic:

```
select Date,Env from read_csv($file);
```

with csv:
```
Date,Env
"foo","bar"
```

and us running:
```
                            (0..stmt.column_count())
                                .map(|i| {
                                    let logical_type = stmt.column_logical_type(i);
                                    if logical_type.is_invalid() {
                                        None
                                    } else {
                                        logical_type.get_alias()
                                    }
                                })
                                .collect::<Vec<
```
on first row (note that i'm using is_invalid here and otherwise we would have fetched get_alias directly and got a panic


diegoimbert:
I'd add that I think that the real problem is that invalid logical type id shouldn't be returned from that query, but it's still nice to be able to check if the id is valid for worst case scenario

mlafeldt:
Good point. I changed LogicalTypeId to be non_exhaustive and return Invalid for unsupported types.

PTAL: https://github.com/duckdb/duckdb-rs/pull/629

PR #631, opened 2025-11-20
Upgrade to arrow 57

mjgarton:
Upgrade the version of arrow from 56 to 57.


mlafeldt:
Thanks for the PR.

While this update may seem benign, it would also require us to bump the MSRV to `1.85.1` due to https://github.com/apache/arrow-rs/blob/57.0.0/Cargo.toml#L85

However, [we committed](https://github.com/duckdb/duckdb-rs?tab=readme-ov-file#rust-version-compatibility) to not change that version in patch releases (1.4.3 would be [next](https://duckdb.org/release_calendar)).

Is there anything in particular you need from Arrow v57?

PR #652, opened 2025-12-18
Remove legacy duckdb_entrypoint macro

mlafeldt:
The old extension API is superseded by `duckdb_entrypoint_c_api`, which uses DuckDB's modern C extension API with proper version negotiation and error handling.


samansmink (approved):
looks good to me! the legacy entrypoint macro was never used in our rust extension template so let's rip it out!

PR #657, opened 2025-12-28
Add logical_type() to vtab::Value

marksomething:
Intending this to be a toe-in-the-water for building out `vtab::Value` (maybe should be renamed `ValueHandle` for consistency with `LogicalTypeHandle`?).

I was trying to create an extension (parsing fixed width binary files...) that needs non-integer parameters so looking to put the plumbing in to make that possible.

### DIFF
--- a/crates/duckdb/src/vtab/value.rs
+++ b/crates/duckdb/src/vtab/value.rs
@@ -1,4 +1,5 @@
-use crate::ffi::{duckdb_destroy_value, duckdb_free, duckdb_get_int64, duckdb_get_varchar, duckdb_value};
+use crate::core::LogicalTypeHandle;
+use crate::ffi::{duckdb_destroy_value, duckdb_free, duckdb_get_int64, duckdb_get_value_type, duckdb_get_varchar, duckdb_value};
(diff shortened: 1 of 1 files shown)


crates/duckdb/src/vtab/value.rs
```diff
@@ -30,6 +31,15 @@ impl Value {
     pub fn to_int64(&self) -> i64 {
         unsafe { duckdb_get_int64(self.ptr) }
     }
+
+    /// Returns a LogicalTypeHandle representing the type of this value.
+    pub fn logical_type(&self) -> LogicalTypeHandle {
+        // SAFETY: The returned LogicalTypeHandle frees allocated memory when dropped.
```
mlafeldt: Not true. `duckdb_get_value_type` returns a borrowed pointer to internal memory owned by the Value, not a new allocation. The C API docs say: "The type itself must not be destroyed". But `LogicalTypeHandle::Drop` calls `duckdb_destroy_logical_type`, causing double-free/corruption when dropped. I suggest returning an owned copy.

Alternatively, if sufficient, implement `logical_type_id` and return `LogicalTypeId`, which doesn't have this ownership problem.

PR #658, opened 2025-12-28
fix(build): prevent cargo cache invalidation from absolute OUT_DIR path

cameronraysmith:


cameronraysmith:
https://github.com/duckdb/duckdb-rs/blob/65e8496fc57098bc896d9fb2ec993b6f0de4e399/crates/libduckdb-sys/build.rs#L140-L144

The absolute `OUT_DIR` path on L142 gets baked into cargo's fingerprint. When artifacts are reused across builds with different `OUT_DIR` values (nix, bazel, or simply cargo clippy vs cargo build in some configurations), cargo treats the path as missing and rebuilds libduckdb-sys from scratch.

History:
- #127 added `rerun-if-changed={}/manifest.json` (relative path) to track the file list
- #163 added `rerun-if-changed=duckdb.tar.gz` which supersedes manifest.json tracking since manifest.json is extracted from the tarball
- #347 moved extraction to `OUT_DIR` for sandbox compliance, changing manifest.json to an absolute path `{out_dir}/{lib_name}/manifest.json`

After #163, L142 became redundant. After #347, it also became non-portable. Deleting L142 fixes the portability issue with no change in rebuild semantics—L144 already covers it.

PR #664, opened 2026-01-20
Fix silent streaming error handling

utay:
When `duckdb_stream_fetch_chunk` returns null, it could indicate either end-of-stream or an error, but the code previously returned `None` in both cases, silently swallowing errors. This PR checks `duckdb_result_error` to distinguish between the two and returns `Result<Option<...>>` so callers can handle streaming failures.

### DIFF
--- a/crates/duckdb/src/arrow_batch.rs
+++ b/crates/duckdb/src/arrow_batch.rs
@@ -1,6 +1,8 @@
+use std::iter::FusedIterator;
+
 use super::{
-    arrow::{datatypes::SchemaRef, record_batch::RecordBatch},
     Result, Statement,
+    arrow::{datatypes::SchemaRef, record_batch::RecordBatch},
 };
 
 /// A handle for the resulting RecordBatch of a query.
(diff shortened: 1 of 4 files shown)


crates/duckdb/src/arrow_batch.rs
```diff
@@ -59,9 +59,18 @@ impl<'stmt> ArrowStream<'stmt> {
...
 
     fn next(&mut self) -> Option<Self::Item> {
-        Some(RecordBatch::from(&self.stmt?.stream_step(self.get_schema())?))
+        let stmt = self.stmt?;
+        match stmt.stream_step(self.get_schema()) {
+            Ok(Some(array)) => Some(Ok(RecordBatch::from(&array))),
+            Ok(None) => None,
+            Err(e) => {
+                // Clear the statement to prevent further iteration after error
+                self.stmt = None;
```
mlafeldt: No test verifies iterator yields `None` here. Should add `assert!(stream.next().is_none())`.

mlafeldt (changes requested):
Thanks!

Good fix and worth the breaking change. 👍

Left some comments for you to address before merging.

PR #668, opened 2026-01-23
Add prepared_schema() to get Arrow schema without executing statement

fvaleye:
# Summary
Adds a new `prepared_schema()` method that extracts the Arrow Schema from a prepared statement without executing it.

When using `stream_arrow()` for streaming query results, the caller must provide the schema upfront. 
Currently, the only way to obtain the schema is to execute the query first, which forces a pattern like:
```rust
// Current workaround: prepare and execute twice 
let mut stmt1 = conn.prepare(sql)?;
stmt1.execute([])?; 
let schema = stmt1.schema();
drop(stmt1); 

let mut stmt2 = conn.prepare(sql)?;  // Second preparation        
let stream = stmt2.stream_arrow([], schema)?;
```                                                                                                                                
                                                                                                                                                                                
This doubles the preparation cost (parsing, planning, optimization) for every streaming query.
With `prepared_schema()`, the schema can be obtained directly from the prepared statement:                                                                                                                                                                                                                                                           
```rust
// New: single preparation                                                                                                                                                            
let mut stmt = conn.prepare(sql)?;                                                                                                                                                    
let schema = stmt.prepared_schema()?.unwrap();                                                                                                                                        
let stream = stmt.stream_arrow([], schema)?;           
```

### DIFF
--- a/crates/duckdb/src/inner_connection.rs
+++ b/crates/duckdb/src/inner_connection.rs
@@ -6,6 +6,11 @@ use std::{
     sync::{Arc, Mutex},
 };
 
+use arrow::{
+    datatypes::{Schema, SchemaRef},
+    ffi::FFI_ArrowSchema,
+};
+
 use super::{ffi, Appender, Config, Connection, Result};
 use crate::{
     error::{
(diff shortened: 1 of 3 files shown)


mlafeldt:
Thanks for putting this together.

With #807 landing, `stream_arrow` no longer requires callers to pass an Arrow schema, so the motivating use case for `prepared_schema` should go away. I'm going to close this as superseded by #807.

If there is still a need to inspect a prepared statement's Arrow schema before execution, please call it out, and we can discuss it as a distinct API.

PR #676, opened 2026-01-30
Build out vtab::Value to enable more diverse parameters to Table Functions

marksomething:
Extend `vtab::Value` to support simple readers, `to_vec()`, `logical_type_id()`, and `is_null()`, this builds up the ability to use more parameter types in table functions and in the case of numbers, get a specific type. Implemented these simple types as a macro for DRY.


mlafeldt:
Thanks! Merged most of this as part of https://github.com/duckdb/duckdb-rs/pull/747

PR #681, opened 2026-02-06
Use Rust 2024 edition

yutannihilation:
Now that the MSRV is bumped to 1.85, I believe duckdb-rs should have no problem to use Rust 2024.

This pull request updates the `edition` to 2024, and introduces necessary changes due to the migration.

- Even in a function that marked as unsafe, unsafe blocks are required
- specifying `ref` in non-move mode  is not allowed (cf. <https://doc.rust-lang.org/reference/patterns.html#r-patterns.ident.binding.mode-limitations.edition2024>)
- `cargo fmt` formats a bit differently

### DIFF
--- a/crates/duckdb/src/vtab/arrow.rs
+++ b/crates/duckdb/src/vtab/arrow.rs
@@ -45,23 +45,23 @@ pub struct ArrowVTab;
 
 unsafe fn address_to_arrow_schema(address: usize) -> FFI_ArrowSchema {
     let ptr = address as *mut FFI_ArrowSchema;
-    *Box::from_raw(ptr)
+    unsafe { *Box::from_raw(ptr) }
 }
 
 unsafe fn address_to_arrow_array(address: usize) -> FFI_ArrowArray {
     let ptr = address as *mut FFI_ArrowArray;
-    *Box::from_raw(ptr)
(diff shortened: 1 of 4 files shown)


yutannihilation:
Wow, great! Sorry for my ignorance...

Let me correct one more thing; it seems bindgen insists it already support Rust 2024.

> Unsafe bitfield raw getters and setters (line 39 in your case) are addressed in https://github.com/rust-lang/rust-bindgen/pull/3124. For other cases, bindgen's --wrap-unsafe-ops option should do the work.  
(https://github.com/rust-lang/rust-bindgen/issues/3147#issuecomment-2676683908)

According to this comment, to support Rust 2024, we should

- Update bindgen to newer than v0.72.0 because https://github.com/rust-lang/rust-bindgen/pull/3124 was merged at the version
- Use `.wrap_unsafe_ops(true)` to wrap other unsafe operations

But, I couldn't make this work on my local. I tried these two and ran `upgrade.sh`, but the generated bindings were the same. Maybe I miss something.

mlafeldt:
Interesting. I think we could also add `.wrap_unsafe_ops(true)` to the builder chain in `build.rs` after bumping the bindgen version. That said, `#[allow]` is simpler IMO, and I went with that instead.

Thanks for the effort here!

PR #682, opened 2026-02-12
Allow binding Date32 and Time64

ultrabear:
These ones seemed straightforward to add, so I threw something together


ultrabear:
Noticed another PR that was working on Appender (adding Decimal), Appender does already support Date32 and Time64

PR #701, opened 2026-03-09
Update DuckDB to v1.5.0

mlafeldt:
Need to tag `v1.10500.0` after merging this.

Follow-up task: change `upgrade.sh` for the new versioning


Maxxen:
Ah yeah, I think @taniabogatsch ran into something similar for the go-driver and had to link something extra but I don't remember completely

PR #732, opened 2026-03-26
Enable static DuckDB extensions via Cargo in checkouts

### LINKED ISSUE #461: Add the option to statically link extensions and disable further extension installation
By default, DuckDB will install and load extensions at run-time. In environments that do not have access to the Internet, or in environments where extension loading must be strictly controlled for cybersecurity or compliance, DuckDB can 1) use a local extension repository, or 2) it can be compiled with the extensions statically linked.

1. The local extension repository can be set using the following DuckDB statement:
    ```
    set custom_extension_repository = 'my/path';
    ```
    The easiest way to seed this is to run DuckDB, install the extensions you want, then set the DuckDB extensions cache (e.g., `~/duckdb/extensions`) to the extension repository path.
2. The other option is to compile DuckDB from source and statically link the extensions. From the DuckDB [documentation on building extensions](https://duckdb.org/docs/stable/dev/building/building_extensions.html):
    ```
    git clone https://github.com/duckdb/duckdb.git
    cd duckdb
    git checkout v1.2.1
    CORE_EXTENSIONS='autocomplete;httpfs;icu;json;tpch' DISABLE_EXTENSION_LOAD=1 GEN=ninja make
    ```
    Furthermore, the `DISABLE_EXTENSION_LOAD=1` will prevent the run-time installation of any additional extensions.

    Running DuckDB and reporting the extensions shows they are statically linked, and additional extensions cannot be loaded:
    ```
	% ./build/release/duckdb
	v1.2.1 8e52ec4395
	Enter ".help" for usage hints.
	Connected to a transient in-memory database.
	Use ".open FILENAME" to reopen on a persistent database.
	D SELECT * FROM duckdb_extensions();
	┌──────────────────┬─────────┬───────────┬───┬───────────────────┬───────────────────┬────────────────┐
	│  extension_name  │ loaded  │ installed │ … │ extension_version │   install_mode    │ installed_from │
	│     varchar      │ boolean │  boolean  │   │      varchar      │      varchar      │    varchar     │
	├──────────────────┼─────────┼───────────┼───┼───────────────────┼───────────────────┼────────────────┤
	│ arrow            │ false   │ false     │ … │                   │ NULL              │                │
	│ autocomplete     │ true    │ true      │ … │ v1.2.1            │ STATICALLY_LINKED │                │
	│ aws              │ false   │ false     │ … │                   │ NULL              │                │
	│ azure            │ false   │ false     │ … │                   │ NULL              │                │
	│ core_functions   │ true    │ true      │ … │ v1.2.1            │ STATICALLY_LINKED │                │
	│ delta            │ false   │ false     │ … │                   │ NULL              │                │
	│ excel            │ false   │ false     │ … │                   │ NULL              │                │
	│ fts              │ false   │ false     │ … │                   │ NULL              │                │
	│ httpfs           │ true    │ true      │ … │ 85ac466           │ STATICALLY_LINKED │                │
	│ iceberg          │ false   │ false     │ … │                   │ NULL              │                │
	│ icu              │ true    │ true      │ … │ v1.2.1            │ STATICALLY_LINKED │                │
	│ inet             │ false   │ false     │ … │                   │ NULL              │                │
	│ jemalloc         │ false   │ false     │ … │                   │ NULL              │                │
	│ json             │ true    │ true      │ … │ v1.2.1            │ STATICALLY_LINKED │                │
	│ motherduck       │ false   │ false     │ … │                   │ NULL              │                │
	│ mysql_scanner    │ false   │ false     │ … │                   │ NULL              │                │
	│ parquet          │ true    │ true      │ … │ v1.2.1            │ STATICALLY_LINKED │                │
	│ postgres_scanner │ false   │ false     │ … │                   │ NULL              │                │
	│ shell            │ true    │ true      │ … │ v1.2.1            │ STATICALLY_LINKED │                │
	│ spatial          │ false   │ false     │ … │                   │ NULL              │                │
	│ sqlite_scanner   │ false   │ false     │ … │                   │ NULL              │                │
	│ tpcds            │ false   │ false     │ … │                   │ NULL              │                │
	│ tpch             │ true    │ true      │ … │ v1.2.1            │ STATICALLY_LINKED │                │
	│ ui               │ false   │ false     │ … │                   │ NULL              │                │
	│ vss              │ false   │ false     │ … │                   │ NULL              │                │
	├──────────────────┴─────────┴───────────┴───┴───────────────────┴───────────────────┴────────────────┤
	│ 25 rows                                                                         9 columns (6 shown) │
	└─────────────────────────────────────────────────────────────────────────────────────────────────────┘
	D INSTALL sqlite_scanner;
	Permission Error:
	Installing external extensions is disabled through a compile time flag
	D INSTALL sqlite_scanner;
    ```

I would like `duckdb-rs` to support this behaviour through `Cargo.toml` features.

It doesn't seem easy to do with the current approach of using the [cc-rs](https://crates.io/crates/cc) crate in the `crates/libduckdb-sys/build.rs` file. It would likely be easier if the `build.rs` file used [cmake](https://crates.io/crates/cmake) or [ninja-writer](https://crates.io/crates/ninja-writer) instead. An attempt has already been made by @njaard to [change `build.rs` to use `cmake`](https://github.com/duckdb/duckdb-rs/pull/387/files#diff-91c73d4bb9ef27d642dd03e1ff7b954455970b6849eaf3acdddf6f35dbb87581), but it was never merged.

 Is there any interest in supporting this feature?

mlafeldt:
This PR adds an experimental `bundled-cmake` build path for duckdb-rs checkouts. It keeps the existing `bundled` feature unchanged, and adds a higher-priority bundled backend that builds DuckDB through upstream CMake when `bundled-cmake` is enabled.

Only in-tree static extensions such as `icu` are supported for now, but there's a path toward future out-of-tree extension support. In fact, I think using CMake is the only realistic path toward out-of-tree extensions. Features like `sqlite_scanner` are already part of DuckDB's checked-in extension configs. Reusing that upstream mechanism is much more maintainable than inventing a parallel extension build/link system in Rust. *Let DuckDB build DuckDB.*

*Update: For out-of-tree extensions, see https://github.com/duckdb/duckdb-rs/pull/734*

Although this PR only targets checkout builds, the CMake backend is structured so that a future crates.io-friendly variant could reuse the same backend logic after downloading the DuckDB sources (`duckdb.tar.gz` does not contain the full source tree).

Fixes #461


mlafeldt:
Merging as an experimental addition, subject to change.

PR #734, opened 2026-03-31
Support static linking of out-of-tree extensions

mlafeldt:
Based on #732


mlafeldt:
Let's revisit once there's actual demand for this feature.

bradhe:
Hey just wanted to throw my support behind this! We're embedding DuckDB into a CLI to query Iceberg tables client-side and I'm looking for a solution to avoid downloading the iceberg, httpfs, parquet, etc., extensions at runtime without doing something completely unnatural 😄

PR #752, opened 2026-04-30
feat: expose duckdb_result_error_type() to populate ErrorCode in DuckDBFailure

### LINKED ISSUE #679: Detecting HTTP 404 (Not Found) responses
First of all, thanks for maintaining this crate 🚀 

I wanted to check something with people that might be more familiar with inner workings of this crate (or duckdb for that matter):

I have a use-case where I export data from Snowflake to S3, and then ingest that data into in-memory duckdb, using ducdkb-rs crate.

Everything is fine and all, but I am not sure what is the best way to handle scenario where there is no data in original Snowflake table: in that case Snowflake does not create empty parquet file, it doesn't do anything really.

If that happens, currently I get an `Error` result, and I figured out from error messages that it it kinda looks like something this f-on should return true for:
```rust
fn is_http_404_error(msg: &str) -> bool {
    let msg = msg.to_lowercase();
    msg.starts_with("http error") && msg.ends_with("404 (not found).")
}
```
As far as I remember, the actual instance is `duckdb::Error::DuckDBFailure`, not sure about `ErrorCode` though.

I have a test suite that started failing on this scenario when I bumped version of duckdb-rs from 1.3.2 to 1.4.4, and the message that I am getting now is `HTTP Error: Unable to connect to URL...`

Is there a more robust way to detect this case (basically 404 from aws), maybe some int representing specific `duckdb_state` or error code or anything like that? Is this something I should search in duckdb GH for (or maybe even AWS extension, or duckdb-httpfs)?

Thanks!

EdanStasiuk:
## Summary

Fixes https://github.com/duckdb/duckdb-rs/issues/679. Currently every `DuckDBFailure` error carries `ErrorCode::Unknown` regardless of what actually went wrong. This is because duckdb-rs only reads a generic success/failure integer from the C API and never calls `duckdb_result_error_type()`, which has been available since DuckDB v0.10 and returns the actual error type.

This PR adds `ErrorCode::Http` and a full mapping from all `duckdb_error_type` C constants to Rust `ErrorCode` variants, then calls `duckdb_result_error_type()` when constructing `DuckDBFailure` errors from a `duckdb_result`. This allows users to match on specific error codes instead of parsing error message strings, which is fragile and breaks between DuckDB versions.

After this change, errors from `execute_streaming` and `execute_extracted_statement` will carry the correct code. For example, `ErrorCode::Http` for HTTP fetch failures, `ErrorCode::ConstraintViolation` for constrain violations, and `ErrorCode::NotFound` for missing tables.

## Tests

All test cases pass except for `types::test::test_dynamic_type`, which is already failing on the main branch.

### DIFF
--- a/crates/duckdb/src/error.rs
+++ b/crates/duckdb/src/error.rs
@@ -297,6 +297,23 @@ pub fn result_from_duckdb_arrow(code: ffi::duckdb_state, mut out: ffi::duckdb_ar
 
 #[cold]
 #[inline]
+pub unsafe fn result_from_duckdb_result(result: *mut ffi::duckdb_result) -> crate::Error {
+    unsafe {
+        let ffi_error = ffi::Error::from_result(result);
+        let msg = {
+            let c_err = ffi::duckdb_result_error(result);
+            if (c_err).is_null() {
+                None
+            } else {
+                Some(CStr::from_ptr(c_err).to_string_lossy().into_owned())
+            }
+        };
+        crate::Error::DuckDBFailure(ffi_error, msg)
+    }
+}
+
+#[cold]
+#[inline]
 pub fn result_from_duckdb_extract(
     num_statements: ffi::idx_t,
     mut extracted: ffi::duckdb_extracted_statements,
--- a/crates/duckdb/src/inner_connection.rs
+++ b/crates/duckdb/src/inner_connection.rs
@@ -10,7 +10,7 @@ use super::{Appender, Config, Connection, Result, ffi};
 use crate::{
     error::{
         Error, result_from_duckdb_appender, result_from_duckdb_arrow, result_from_duckdb_extract,
-        result_from_duckdb_prepare,
+        result_from_duckdb_prepare, result_from_duckdb_result,
     },
     raw_statement::RawStatement,
     statement::Statement,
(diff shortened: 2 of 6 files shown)


EdanStasiuk:
As this PR currently stands, all relevant unit tests are passing except for `test::test_interrupt`, which is caused by my re-executing via [`duckdb_execute_prepared`](https://github.com/duckdb/duckdb-rs/pull/752/changes#diff-35e96f4ccb65761cfbd21a3289b0c271cb4980aab9f4f7915ad843324379d0daR305) after `duckdb_execute_prepared_arrow` fails inside `execute()`. When a query is interrupted the connection is in a bad state and the second execution hangs.

The way I see it, there's 3 way to address this:
1. Check the arrow error message before re-executing, and if it contains "INTERRUPT" skip the re-execution and return immediately. This fixes the test but is fragile since it's string matching on error messages, which is exactly what this PR is trying to move away from.
2. Leave the arrow path untouched and revert `test_notnull_constraint_error` [back to asserting `ErrorCode::Unknown`](https://github.com/duckdb/duckdb-rs/pull/752/changes#diff-b0e1bf42d464b4842060b8de6842557c03957fda40b1b72b17cea2634dec3335L1105-L1106). Typed errors would still work for `execute_streaming` and `execute_extracted_statement`, which covers the HTTP error case from the original issue.
3. Migrate `execute()` off `duckdb_execute_prepared_arrow` entirely and onto `duckdb_execute_prepared`. This would give typed errors without any re-execution hack, and also addresses the fact that `duckdb_execute_prepared_arrow` is marked as deprecated in the C API. However this requires replacing all the data reading code in `RawStatement` that currently depends on the `duckdb_arrow` handle, which looks to be a significant rewrite and out of scope for this PR.

Happy to go in whichever direction the maintainers prefer.

EdanStasiuk:
Un-drafting the PR with option 2, as described previously. Happy to make any changes

PR #753, opened 2026-05-01
libduckdb-sys: expose DEP_DUCKDB_INCLUDE for downstream C/C++ shims

frhack:
## What

Add `links = "duckdb"` to `libduckdb-sys`'s manifest and emit `cargo:include=<path>` from each backend so downstream crates that compile their own C/C++ code can read the resolved DuckDB include directory from `DEP_DUCKDB_INCLUDE` in their own build scripts.

## Why

Crates that ship native extension code (e.g. an `OptimizerExtension` shim that `#include "duckdb.hpp"`) need access to libduckdb-sys's DuckDB headers at their own build time.

Today they have to either:
- Glob `target/<profile>/build/libduckdb-sys-<hash>/out/duckdb/src/include` from their `build.rs`, or
- Vendor a copy of the DuckDB headers into their own crate.

Both are fragile: the glob breaks across cargo versions and is awkward in cross-builds; vendoring drifts from the bundled DuckDB ABI on every `libduckdb-sys` upgrade.

The standard Cargo mechanism for this is the `links` directive plus `cargo:KEY=VALUE` metadata. With this PR, downstream code can simply:

```rust
// downstream build.rs
fn main() {
    let inc = std::env::var("DEP_DUCKDB_INCLUDE").unwrap();
    cc::Build::new()
        .cpp(true)
        .file("cpp/extension_shim.cpp")
        .include(&inc)
        .compile("my_extension_shim");
}
```

This unblocks downstream crates from registering things like `duckdb::OptimizerExtension` from their own native code without resorting to the C API (which the issue at duckdb/duckdb#19093 is still working out).

## Changes

- `crates/libduckdb-sys/Cargo.toml`: add `links = "duckdb"` so build-script metadata flows to downstream `DEP_DUCKDB_*` env vars.
- `crates/libduckdb-sys/build_bundled_cc.rs`: emit `cargo:include=<OUT_DIR>/duckdb/src/include`.
- `crates/libduckdb-sys/build_bundled_cmake.rs`: emit `cargo:include=<duckdb-sources>/src/include`.
- `crates/libduckdb-sys/build.rs` (the `build_linked` path): emit `cargo:include=...` derived from the resolved `HeaderLocation` — `DUCKDB_INCLUDE_DIR`, `DUCKDB_LIB_DIR`, vcpkg, pkg-config, or download dir.

Wrapper-only mode (system DuckDB found via pkg-config without an explicit include path) intentionally skips the emission — there's no single directory to publish; the system header is already on the default search path, and downstream shims fall through to the same default include resolution.

## Behaviour change

None for existing Rust consumers — the only public-surface change is a new `links` directive, which doesn't affect compilation of crates that don't run their own `build.rs` against duckdb headers.

## Validation

The companion downstream project that motivated this PR (`dbfy` — embedded SQL federation engine, [github.com/frhack/dbfy](https://github.com/frhack/dbfy)) ships an `OptimizerExtension` shim in bundled mode (`--features duckdb`) that needs these headers. Its `build.rs` currently uses the target-tree glob hack as a documented workaround; once this PR is in a release, the shim can switch to the clean `DEP_DUCKDB_INCLUDE` path.

Happy to add a small test or example exercising the metadata if useful.

### DIFF
--- a/crates/libduckdb-sys/build_bundled_cmake.rs
+++ b/crates/libduckdb-sys/build_bundled_cmake.rs
@@ -37,7 +37,14 @@ pub fn main(_out_dir: &str, out_path: &Path) {
     println!("cargo:rerun-if-env-changed=CMAKE_CXX_COMPILER_LAUNCHER");
     println!("cargo:rerun-if-env-changed=MACOSX_DEPLOYMENT_TARGET");
 
-    write_bindings(&source_dir.join("src/include"), out_path);
+    let include_path = source_dir.join("src/include").canonicalize().unwrap();
+    write_bindings(&include_path, out_path);
+
+    // Publish the include directory so downstream crates that compile
+    // their own C/C++ code can read it from `DEP_DUCKDB_INCLUDE`.
+    // See the matching emission in build_bundled_cc.rs.
+    println!("cargo:include={}", include_path.display());
+
     if let Some(configs) = env_var("DUCKDB_EXTENSION_CONFIGS") {
         if !configs.trim().is_empty() {
             panic!(
--- a/crates/libduckdb-sys/Cargo.toml
+++ b/crates/libduckdb-sys/Cargo.toml
@@ -8,6 +8,12 @@ homepage = { workspace = true }
 keywords = { workspace = true }
 readme = { workspace = true }
 build = "build.rs"
+# `links = "duckdb"` lets the build script publish metadata
+# (`cargo:include`, `cargo:lib_dir`) that downstream crates can read
+# via `DEP_DUCKDB_*` environment variables in their own build scripts.
+# Required for crates that need to compile their own C/C++ code that
+# `#include "duckdb.hpp"` against the bundled headers.
+links = "duckdb"
 categories = ["external-ffi-bindings", "database"]
 description = "Native bindings to the libduckdb library, C API"
 exclude = ["duckdb-sources"]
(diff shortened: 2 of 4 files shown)


crates/libduckdb-sys/build_bundled_cmake.rs
```diff
@@ -37,7 +37,14 @@ pub fn main(_out_dir: &str, out_path: &Path) {
     println!("cargo:rerun-if-env-changed=CMAKE_CXX_COMPILER_LAUNCHER");
     println!("cargo:rerun-if-env-changed=MACOSX_DEPLOYMENT_TARGET");
 
-    write_bindings(&source_dir.join("src/include"), out_path);
+    let include_path = source_dir.join("src/include");
```
mlafeldt: This is currently relative but should be absolute so downstream build scripts can actually use it.

```suggestion
    let include_path = source_dir.join("src/include").canonicalize().unwrap();
```

mlafeldt (commented):
Thanks!

Happy to merge this once you address the mentioned path bug and format your code with cargo fmt.

mlafeldt:
@frhack Still interested in getting this in? We could ship this as part of DuckDB 1.5.3 this week.

frhack:
Hi,
yes I'm still interested. This evening I will fix the PR

Thanks

Francesco

frhack:
> Thanks!
> 
> Happy to merge this once you address the mentioned path bug and format your code with cargo fmt.

Thanks! I addressed both review points:

 - `bundled-cmake` now canonicalizes the include path before publishing `cargo:include`, so downstream `DEP_DUCKDB_INCLUDE` is absolute.
 - Ran `cargo fmt` and amended the PR commit.

I also verified `cargo fmt -- --check` and `git diff --check`. A local `cargo check -p libduckdb-sys --no-default-features --features bundled-cmake` gets past build-script compilation but cannot complete in my fresh clone because the `duckdb- sources` submodule is not initialized.

mlafeldt (approved):
Thanks!

Pushed a small follow-up fix and now it's ready. 🤙

Issue #680, opened 2026-02-06
Add bind_parameter support for "container" types


aljazerzen:
Related to #619

Bind parameter does not support `Value::Struct`, `Value::Array`, `Value::Map` or `Value::Union`:

```rs
    // generic because many of these branches can constant fold away.
    fn bind_parameter<P: ?Sized + ToSql>(&self, param: &P, col: usize) -> Result<()> {
        let value = param.to_sql()?;

        let ptr = unsafe { self.stmt.ptr() };
        let value = match value {
            ToSqlOutput::Borrowed(v) => v,
            ToSqlOutput::Owned(ref v) => ValueRef::from(v),
        };
        // TODO: bind more
        let rc = match value {
            ...
            _ => unreachable!("not supported: {}", value.data_type()),
        };
        ...
```

A test:

```rs
use duckdb::types::Value;
use duckdb::{params, Connection, Result};

fn main() -> Result<()> {
    let conn = Connection::open_in_memory()?;

    conn.execute("CREATE TABLE test (id INTEGER, numbers INTEGER[])", [])?;

    let list_value = Value::List(vec![Value::Int(1), Value::Int(2), Value::Int(3)]);
    conn.execute("INSERT INTO test VALUES (?, ?)", params![1, &list_value])?;

    Ok(())
}
```

mlafeldt:
Hey @aljazerzen,

#749 just landed as a stopgap: instead of panicking on unsupported Value variants (List, Array, Struct, Map, Union, Enum), bind/append now return `Error::ToSqlConversionFailure` with a clear "binding/appending X is not yet supported" message. So no more crashes, but the actual feature you want is still missing.

Want to take a stab at it? The path forward is wiring real conversion in `value_ref_from_value` (and the appender's `bind_parameter`) for one variant at a time. `List` is probably the most natural starting point. Happy to review. Please keep PRs small and scoped so they're easy to land. 🚀

mlafeldt:
Oh, I totally forgot about #643 by @ozars, which already implements `List`. It's been open for a while (my bad) and could use a nudge / review / rebase to get across the line.

aljazerzen:
Yes, I can take a stab. I will probably start anew and port `List` again, minimally, and then other types. It might take me some time, as this is not a top priority anymore.

aljazerzen:
re-reading these messages looks like two parrots with knives:

> I can take a stab at ...

> Want to take a stab at it? 

> Yes, I can take a stab.

[...]

PR #757, opened 2026-05-11
Implement binding for List

aljazerzen:
Resolves a part of #680
Susersedes #643

I recomment review commit-by-commit:

- **Add Type::name() and Value::data_type_name()**
- **Add ToSqlOutput::BorrowedValue variant**
- **Bind and append List values**

## Main problem

`Value::List` is not cheaply convertible into `ValueRef::List`.

Before this PR, binding (and appending) converted all values to `ValueRef`,
and then matched on the variant to call appropriate `ffi::duckdb_bind_xxx`.

For container types, we cannot cheaply convert between `Value` and `ValueRef`,
so I've added two binding paths: `value_to_duckdb` and `value_ref_to_duckdb`.

Both convert to `ffi::duckdb_value`, which is then passed to
`ffi::duckdb_bind_value`/`ffi::duckdb_append_value`.

These two helpers do also support converting all other types, but they always
use the faster path of calling specialized ffi functions. 

## Why add `ToSqlOutput::BorrowedValue` variant?

Because we need to call `ToSql(Value)` and return `&Value`, not `ValueRef`.

In essence, the problem is that there are two dimensions in which `Value`
and `ValueRef` differ:
- owned / borrowed,
- statically / dynamically typed.

For most types except container types, both of these distinctions are irrelevant.
But for container types (e.g `List`), there can be 4 possible representations:

- owned dynamic: `Vec<Value>`, (this is `Value`)
- owned static: `arrow::ListArray`,
- borrowed dynamic: `&[Value]`,
- borrowed static: `&arrow::ListArray`.

When binding, `ToSql` needs a path from `owned dynamic` to `borrowed dynamic`.

PR #643 gets around this by adding a new variant to ListType, but I think that
is messy because it mixes static/dynamic distinction with arrow list type impls.

### DIFF
--- a/crates/duckdb/src/appender/mod.rs
+++ b/crates/duckdb/src/appender/mod.rs
@@ -4,7 +4,8 @@ use std::{ffi::c_void, fmt, os::raw::c_char};
 use crate::{
     Error,
     error::result_from_duckdb_appender,
-    types::{ToSql, ToSqlOutput},
+    statement::{value_ref_to_duckdb, value_to_duckdb},
+    types::{ToSql, ToSqlOutput, Value},
 };
 
 /// Appender for fast import data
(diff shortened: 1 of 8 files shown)


aljazerzen:
@mlafeldt What's happening here? This is blocking me and it seem a bit rude to ask for a PR and then not review it.

nimdeveloper:
I've created a new crate, implemented almost many of types, you can check it here
https://github.com/nimdeveloper/better-duck
It may help

aljazerzen:
This appears to be too out of sync. @mlafeldt It is very unprofessional to ask for contributions and then not even review or reply. I've spent a day of work on this.

PR #758, opened 2026-05-16
Change LogicalTypeHandle::set_alias to take &mut self

vergenzt:
Since it mutates the underlying data, no? Leaving the signature with an immutable `&self` borrow means the compiler will let multiple threads / calls write to the data at the same time, which is not great. Requiring a mutable borrow guarantees exclusivity, avoiding data races (assuming manual `unsafe` guarantees are maintained).

This may be a (minor) breaking change; e.g. someone may have to make a change like the following:

```diff
-let handle = ...;
+let mut handle = ...;
 handle.set_alias(...);
 ...
```

However IMO it's worth it for more accurate contract description.

### DIFF
--- a/crates/duckdb/src/core/logical_type.rs
+++ b/crates/duckdb/src/core/logical_type.rs
@@ -365,7 +365,7 @@ impl LogicalTypeHandle {
     }
 
     /// Set the alias of the logical type.
-    pub fn set_alias(&self, alias: &str) {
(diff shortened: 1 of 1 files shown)


mlafeldt (approved):
`LogicalTypeHandle` is neither `Send` nor `Sync`, so safe Rust already prevents moving it to another thread or sharing references to it across threads. This change still makes sense from an API perspective as it communicates mutation and enforces exclusive access. 👍

Issue #786, opened 2026-06-13
Problems compiling libduckdb-sys


huhlig:
Newer MSVC toolchains expose two issues when building bundled DuckDB on Windows:

1. MSVC 19.40 removed the non-standard stdext::checked_array_iterator extension. DuckDB's bundled fmt header checks only whether _SECURE_SCL is defined, so debug/test-profile builds can select a branch that references a removed type.
2. DuckDB's Windows file-lock diagnostics call Restart Manager APIs, so final binaries that link bundled DuckDB also need Rstrtmgr.lib.

Patch the extracted bundled fmt header after unpacking and emit the Restart Manager link directive for Windows builds.

huhlig:
```patch
diff --git a/crates/libduckdb-sys/build.rs b/crates/libduckdb-sys/build.rs
index db5ea8a..6bc0a2a 100644
--- a/crates/libduckdb-sys/build.rs
+++ b/crates/libduckdb-sys/build.rs
@@ -81,8 +81,28 @@ mod build_bundled {
         archive.unpack(out_dir).expect("archive");
     }
 
+    fn patch_msvc_removed_checked_iterator(out_dir: &str) {
+        let format_header = format!("{out_dir}/duckdb/third_party/fmt/include/fmt/format.h");
+        let Ok(source) = std::fs::read_to_string(&format_header) else {
+            return;
+        };
+
+        // MSVC 19.40 removed the non-standard stdext::checked_array_iterator
+        // extension. DuckDB's bundled fmt still checks only whether _SECURE_SCL
+        // is defined, so debug builds on newer MSVC try to use a removed type.
+        let patched = source.replace(
+            "#ifdef _SECURE_SCL\n// Make a checked iterator to avoid MSVC warnings.",
+            "#if defined(_SECURE_SCL) && _SECURE_SCL && (!defined(_MSC_VER) || _MSC_VER < 1940)\n// Make a checked iterator to avoid MSVC warnings.",
+        );
+
+        if patched != source {
+            std::fs::write(format_header, patched).expect("patching bundled fmt header");
+        }
+    }
+
     pub fn main(out_dir: &str, out_path: &Path) {
         untar_archive(out_dir);
+        patch_msvc_removed_checked_iterator(out_dir);
 
         if !cfg!(feature = "bundled") {
             // This is just a sanity check, the top level `main` should ensure this.
@@ -169,6 +189,11 @@ mod build_bundled {
         }
         cfg.compile("duckdb");
 
[...]

PR #787, opened 2026-06-17
Enable pre-release testing

mlafeldt:
Add `--sha` override to pin the bundled DuckDB sources to a specific commit, with or without a version bump:

    ./upgrade.sh v1.5.4 --sha 08e34c447b    # planned release, source pinned to commit
    ./upgrade.sh --sha 08e34c447b           # test a commit, no version bump

refs https://github.com/duckdblabs/duckdb-internal/issues/7516


mlafeldt:
CI failure `error C2653: 'stdext': is not a class or namespace name` is caused by #786, which will be fixed upstream. Ignoring.

PR #798, opened 2026-06-29
chore: upgrade reqwest build dependency from 0.12 to 0.13

notmd:
## Summary

Upgrades the `reqwest` build dependency in `libduckdb-sys` from `0.12` to `0.13`.

The only required change beyond the version bump is renaming the `rustls-tls` feature to `rustls`, which was renamed in reqwest 0.13.

The build script usage (`blocking::Client`, `.get().send()`, `io::copy`) is API-compatible between the two versions — no `build.rs` changes were needed.

## Motivation

reqwest 0.13 is the current stable release. Having `libduckdb-sys` pin to 0.12 forces downstream users to compile two versions of reqwest simultaneously, increasing build times and binary size.

## Test plan

- [x] `cargo check` passes on the workspace

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -347,6 +347,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.17.0"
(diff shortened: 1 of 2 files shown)


mlafeldt:
@notmd Please tell Claude to update Cargo.lock

notmd:
Unfortunately, we need to wait for polar to upgrade reqwest to 0.13

mlafeldt:
Replaced reqwest with ureq in #826, so this should no longer be an issue.

PR #801, opened 2026-07-01
Support GEOMETRY type

mlafeldt:
Adds first-class DuckDB `GEOMETRY` support in row decoding, logical type metadata, and value APIs, exposing geometry values as WKB while preserving blob compatibility. Also adds CRS metadata access, geometry coverage for all-types/vtab paths, and appender improvements for schema/catalog selection plus clearer bind-time errors.

Closes #714


crates/duckdb/src/test_all_types.rs
```diff
@@ -19,7 +60,7 @@ fn test_large_arrow_types() -> crate::Result<()> {
 }
 
 fn test_with_database(database: &Connection) -> crate::Result<()> {
-    let excluded = ["time_tz", "time_ns", "bignum", "geometry"];
+    let excluded = ["time_tz", "time_ns", "bignum"];
```
mlafeldt, line 63: Getting smaller and smaller 🤙

Issue #540, opened 2025-07-18
[Feature] Derive `FromSql`, `ToSql`


rm-dr:
I've been using duckdb for a few months, and have written a crate that allows my team to `#[derive]` traits equivalent to `ToSql` and `FromSql`.

Would you consider a PR to integrate these changes into `duckdb-rs` (gated behind a `"derive"` feature), or should I maintain this as a separate project?

**code:** https://git.betalupi.com/Mark/libduck
**crates.io:** https://crates.io/crates/libduck

Issue #329, opened 2024-06-06
`FromSql` is not implemented for `Vec<f64>`


0x1997:
Is there a way to get a `Vec` from an array column?

dvic:
I think you can decode it as an `Value::List`: https://docs.rs/duckdb/latest/duckdb/types/enum.Value.html#variant.List

metasim:
> I think you can decode it as an `Value::List`: https://docs.rs/duckdb/latest/duckdb/types/enum.Value.html#variant.List

Is there a way of doing this without having to iterate over `Vec<Value>` and convert each element into `f64`, thereby creating a copy of the vector? We've a situation where the vectors aren't small.

fotonick:
[Extracting a `List` result](https://github.com/duckdb/duckdb-rs/blob/1096461ea9d6a9632e5cb079860ea47b2de0257b/crates/duckdb/src/types/value_ref.rs#L221) calls [`from_list()`](https://github.com/duckdb/duckdb-rs/blob/1096461ea9d6a9632e5cb079860ea47b2de0257b/crates/duckdb/src/types/value_ref.rs#L307), which just iterates over those values and calls `.to_owned()` on each one, as I think you're trying to avoid doing.

```rust
fn from_list(start: usize, end: usize, idx: usize, values: &ArrayRef) -> Value {
    Value::List(
        (start..end)
            .map(|row| Row::value_ref_internal(row, idx, values).to_owned())
            .collect(),
    )
}
```

fotonick:
To put it in more general terms, if I understand how this all works, every result row is a bunch of `ValueRef`s, which are pointers to memory that duckdb (not duckdb-rs) owns. I would guess that this memory is freed when you finish iterating to the end of a chunk, so you'd better make a copy.

PR #820, opened 2026-07-17
Vector binding, nested containers, and serde STRUCT support

npatsakula:
DuckDB vectors (embeddings) are the `ARRAY` type (`FLOAT[768]`). Reading `ARRAY`/`LIST`/`STRUCT`/`MAP` worked via Arrow, but **binding/appending** them was rejected, forcing SQL casts. This PR wires up binding + ergonomic conversions + a serde bridge. No DuckDB upgrade needed (required C symbols already bound).

## What

- **Container binding** (`types/value_ffi.rs`): recursive `Value -> duckdb_value` for `List`/`Array`/`Struct`/`Map` and any nesting, via `duckdb_bind_value` / `duckdb_append_value`. DuckDB copies children (verified), so an RAII guard frees each built value. `Value::data_type()` for containers (was `todo!()`).
- **Ergonomic vectors**: `ToSql`/`FromSql` for `Vec<f32>`/`Vec<f64>`/`[T; N]` (+ integer vecs). Empty -> error; `NULL` via `Option<T>`. `Vec<u8>` stays `BLOB`.
- **`ndarray` feature** (in `modern-full`): `Array1<f32>`/`Array1<f64>`.
- **`serde` feature** (in `modern-full`): `Struct<T>` wrapper -> `impl<T: Serialize> ToSql` + `FromSql for Struct<T: DeserializeOwned>`. Structs/maps -> `STRUCT`, `Vec` -> `LIST`, unit enums -> `VARCHAR`. Wrapper (not blanket) to dodge coherence overlap with scalar `Serialize` impls.

## Usage

```rust
// vector search
let id: i32 = conn.query_row(
    "SELECT id FROM vecs ORDER BY array_distance(v, ?::FLOAT[3]) LIMIT 1",
    [vec![3.9_f32, 5.0, 6.0]], |r| r.get(0))?;

// serde STRUCT round-trip
#[derive(Serialize, Deserialize)] struct Point { x: i32, y: f64 }
conn.execute("INSERT INTO t VALUES (?)", [Struct(&Point { x: 3, y: 4.5 })])?;
let Struct(p): Struct<Point> = conn.query_row("SELECT p FROM t", [], |r| r.get(0))?;
```

## Caveats

`Value::Enum`/`Value::Union` still unbindable (no tag/type). Borrowed `ValueRef::List` (from Arrow) still rejected — only **owned** container `Value`s use the new path. Arrow appender remains the fast lane for bulk loads.

## Connected Issues

Fixing #540 
Fixing #329

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -870,6 +870,7 @@ dependencies = [
  "rand 0.9.4",
  "rust_decimal",
  "rustyline",
+ "serde",
  "serde_json",
  "strum",
  "tempfile",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -860,6 +860,7 @@ dependencies = [
  "fallible-streaming-iterator",
  "hashlink",
  "libduckdb-sys",
+ "ndarray",
  "num",
  "num-integer",
  "polars",
@@ -1766,6 +1767,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "matrixmultiply"
+version = "0.3.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f607c237553f086e7043417a51df26b2eb899d3caff94e6a67592ff992fedc7"
+dependencies = [
+ "autocfg",
+ "rawpointer",
+]
+
+[[package]]
 name = "memchr"
 version = "2.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


npatsakula:
Hi @mlafeldt! Can I ask you to take a look at this PR? I'm happy to change anything if needed.

PR #821, opened 2026-07-20
Refactor vectors around a native vector seam

mlafeldt:
- Introduce a native vector seam across core vector, data chunk, scalar, and table-function APIs
- Update flat, list, array, struct, writable, lease, resize, and state handling
- Align Arrow interop and nested-vector tests with the new interfaces
- Refresh vector-related examples and callback integrations


mlafeldt:
Scope too big. Will break this up into multiple PRs.

PR #822, opened 2026-07-20
Return errors from Config::set() instead of panicking

arijitroy003:
## Summary

`Config::set()` already returns `Result<()>` but contained two panic paths:

- `CString::new(key).unwrap()` / `CString::new(value).unwrap()` — panics if key or value contains an interior nul byte
- `assert_eq!(state, ffi::DuckDBSuccess)` — panics if `duckdb_create_config` fails

Both are replaced with proper error returns using the existing `Error::NulError` (via the `From<NulError>` impl and `?`) and `Error::DuckDBFailure` variants respectively.

Also includes two minor cleanups:
- Simplify `(enabled as u8).to_string()` to `enabled.to_string()` since `bool` implements `Display` directly with `"true"`/`"false"`
- Fix `http://` to `https://` in workspace `documentation` URL


crates/duckdb/src/config.rs
```diff
@@ -54,7 +54,7 @@ impl Config {
 
     /// enable autoload extensions
     pub fn enable_autoload_extension(mut self, enabled: bool) -> Result<Self> {
-        let value = (enabled as u8).to_string();
+        let value = enabled.to_string();
```
mlafeldt: Although DuckDB's boolean config parser accepts both, this is a behavior change not required by the fix.

mlafeldt (changes requested):
Fix is good, thanks.

Two things before merge:

- Please revert the unnecessary bool cast
- Add a test that exercises the new error behavior

arijitroy003:
Thanks for the review! Pushed a fix:
- Reverted the bool cast change
- Added `test_null_byte_in_key_returns_error` that verifies `Config::set()` returns `Err` instead of panicking when the key contains a null byte

mlafeldt (approved):

mlafeldt:
Thanks!

PR #835, opened 2026-08-28
Add statically bundled httpfs support to bundled-cmake

deem0n:
Adds a checkout-only `httpfs` feature that statically links DuckDB's HTTP(S) and S3 filesystem extension through `bundled-cmake` on Linux and macOS. `LOAD httpfs` works with autoinstall/autoload and external extension loading disabled, without downloading or writing an extension artifact.

- Let CMake discover curl and provide only an OpenSSL prefix hint through pkg-config. Forward curl/OpenSSL linker metadata and the macOS CoreFoundation/Security frameworks required by HTTPFS.
- Preserve the plain `pkg-config` feature for non-bundled discovery and compile HTTPFS helpers only when `httpfs` is enabled.
- Run `test_extension_httpfs` beside the ICU test in Linux/macOS CMake CI, and check CMake without default features. Document build prerequisites, configure-time network access, and unsupported Windows HTTPFS builds.

Validation before updating this PR: [all eight testbed CI jobs passed](https://github.com/deem0n/duckdb-rs/actions/runs/35146401908) on commit `93e51a4179f6caa50592515067de8390873709e5`. This covers Linux/macOS/Windows tests and CMake builds, Linux Clippy with warnings denied, documentation/package checks, AddressSanitizer, and MSRV. Both Linux and macOS executed the static HTTPFS test successfully. [Testbed PR and local validation details](https://github.com/deem0n/duckdb-rs/pull/1).


crates/libduckdb-sys/build_bundled_cmake.rs
```diff
@@ -64,6 +64,10 @@ pub fn main(out_dir: &str, out_path: &Path) {
     configure_macos_deployment_target(&mut config);
     config
         .profile(&cmake_build_type)
+        .define(
+            "OVERRIDE_GIT_DESCRIBE",
+            format!("v{}", duckdb_version_from_pkg_version(env!("CARGO_PKG_VERSION"))),
```
mlafeldt, line 69: This looks unrelated to the httpfs change and should be moved to a separate PR.

I can see the value (shallow submodule checkouts currently report `v0.0.1`), but nightly builds pin sources via `upgrade.sh --sha` without bumping `CARGO_PKG_VERSION`, so they would misreport the release version and autoinstall would fetch release extension binaries against a nightly core.

crates/libduckdb-sys/build.rs
```diff
@@ -25,6 +25,20 @@ fn is_compiler(compiler_name: &str) -> bool {
     std::env::var("CARGO_CFG_TARGET_ENV").is_ok_and(|v| v == compiler_name)
 }
 
+fn duckdb_version_from_pkg_version(pkg_version: &str) -> String {
```
mlafeldt, line 28: This needs a `#[cfg(any(not(feature = "bundled"), feature = "bundled-cmake"))]` to fix clippy/CI.

crates/libduckdb-sys/Cargo.toml
```diff
@@ -21,14 +21,15 @@ edition = { workspace = true }
...
+default = ["vcpkg", "dep:pkg-config"]
 bundled = ["cc"]
 # Warning: experimental feature
 bundled-cmake = ["bundled", "dep:cmake", "dep:which", "parquet"]
-buildtime_bindgen = ["bindgen", "pkg-config", "vcpkg"]
+buildtime_bindgen = ["bindgen", "dep:pkg-config", "vcpkg"]
 json = ["bundled"]
 parquet = ["bundled"]
 autocomplete = ["bundled-cmake"]
+httpfs = ["bundled-cmake", "dep:pkg-config"]
```
mlafeldt, line 32: Please keep the plain name: `httpfs = ["bundled-cmake", "pkg-config"]`

And revert the `dep:` changes above. Switching to `dep:pkg-config` deletes the implicit pkg-config feature. All five `#[cfg(feature = "pkg-config")]` gates in build_linked are now permanently false.

crates/libduckdb-sys/build_bundled_cmake.rs
```diff
@@ -144,9 +152,59 @@ pub fn main(out_dir: &str, out_path: &Path) {
     }
     link_static_library(&lib_dir, &cmake_build_type, "duckdb_static");
     link_system_libs();
+    if cfg!(feature = "httpfs") {
+        emit_httpfs_link_metadata();
+    }
     println!("cargo:lib_dir={}", lib_dir.display());
 }
 
+fn configure_httpfs_dependencies(config: &mut cmake::Config) {
```
mlafeldt, line 161: This fails for me on macOS out of the box and also reimplements parts of CMake. What works:

```diff
diff --git crates/libduckdb-sys/build_bundled_cmake.rs crates/libduckdb-sys/build_bundled_cmake.rs
index 7e9b569f..01c5b3e0 100644
--- crates/libduckdb-sys/build_bundled_cmake.rs
+++ crates/libduckdb-sys/build_bundled_cmake.rs
@@ -159,15 +159,12 @@ pub fn main(out_dir: &str, out_path: &Path) {
 }
 
 fn configure_httpfs_dependencies(config: &mut cmake::Config) {
-    let curl = probe_dependency("libcurl", false);
-    let openssl = probe_dependency("openssl", false);
-
-    config
-        .define("CURL_INCLUDE_DIR", dependency_include_dir(&curl, "curl/curl.h"))
-        .define("CURL_LIBRARY_RELEASE", dependency_library(&curl, "curl"))
-        .define("OPENSSL_INCLUDE_DIR", dependency_include_dir(&openssl, "openssl/ssl.h"))
-        .define("OPENSSL_SSL_LIBRARY", dependency_library(&openssl, "ssl"))
-        .define("OPENSSL_CRYPTO_LIBRARY", dependency_library(&openssl, "crypto"));
+    // Only OpenSSL needs a hint (Homebrew keeps it keg-only). Do not hint curl:
+    // FindCURL locates the sysroot curl itself, and a pkg-config prefix may point
+    // at a different SDK than the compiler, breaking libc++ header ordering.
+    let openssl_prefix = pkg_config::get_variable("openssl", "prefix")
+        .unwrap_or_else(|err| panic!("bundled-cmake httpfs requires openssl through pkg-config: {err}"));
+    config.define("OPENSSL_ROOT_DIR", openssl_prefix);
 }
 
 fn emit_httpfs_link_metadata() {
@@ -182,29 +179,6 @@ fn probe_dependency(name: &str, cargo_metadata: bool) -> pkg_config::Library {
         .unwrap_or_else(|err| panic!("bundled-cmake httpfs requires {name} through pkg-config: {err}"))
 }
 
-fn dependency_include_dir(library: &pkg_config::Library, header: &str) -> PathBuf {
-    library
-        .include_paths
-        .iter()
-        .find(|path| path.join(header).is_file())
-        .cloned()
-        .unwrap_or_else(|| panic!("pkg-config did not report an include directory containing {header}"))
-}
-
-fn dependency_library(library: &pkg_config::Library, name: &str) -> PathBuf {
-    let filenames = [
-        format!("lib{name}.dylib"),
-        format!("lib{name}.so"),
-        format!("lib{name}.a"),
-    ];
-    library
-        .link_paths
-        .iter()
-        .flat_map(|directory| filenames.iter().map(move |filename| directory.join(filename)))
-        .find(|path| path.is_file())
-        .unwrap_or_else(|| panic!("pkg-config did not report a link path containing lib{name}"))
-}
-
 fn cmake_build_type() -> String {
     for var in ["DUCKDB_CMAKE_BUILD_TYPE", "CMAKE_BUILD_TYPE"] {
         if let Some(value) = env_var(var).filter(|v| !v.trim().is_empty()) {
```

crates/duckdb/tests/nightly.rs
```diff
@@ -34,6 +34,51 @@ fn can_install_and_load_httpfs() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+#[test]
+#[cfg(feature = "httpfs")]
```
mlafeldt, line 38: - Gate the pkg-config helpers with `#[cfg(feature = "httpfs")]` since `--no-default-features --features bundled-cmake` no longer compiles.
- This test never runs in CI and doesn't belong in `nightly.rs`. Move it next to `test_extension_icu` and enable `httpfs` in the `cmake` job.
- Don't hardcode `v1.5.5`; derive it from `CARGO_PKG_VERSION`.
- README should list the prerequisites (pkg-config, libcurl/OpenSSL, network at configure time) and say Windows is unsupported.

mlafeldt (changes requested):
Thanks! A curated `httpfs` feature is the right shape, so I'd like to get this in.

Blockers before that: CI is red (dead code under `bundled`), the `dep:pkg-config` change silently disables pkg-config discovery for non-bundled builds, and the build fails out of the box on macOS/Homebrew. The `OVERRIDE_GIT_DESCRIBE` change should go in its own PR. See details and a tested fix for the macOS issue inline.

crates/libduckdb-sys/build_bundled_cmake.rs
deem0n: Removed it for now

crates/libduckdb-sys/build.rs
deem0n: Removing the version override also removed the reason to share that helper with CMake.

  duckdb_version_from_pkg_version() now lives inside build_linked, which already has this gate:

```rust
  #[cfg(not(feature = "bundled"))]
  mod build_linked {
      // ...
      fn duckdb_version_from_pkg_version(...) -> String { ... }
  }
```

crates/libduckdb-sys/Cargo.toml
deem0n: done, as requested

crates/libduckdb-sys/build_bundled_cmake.rs
deem0n: Thanks for hint, fixed

crates/duckdb/tests/nightly.rs
deem0n: 1. Added #[cfg(feature = "httpfs")] to both helpers and their call sites. CI checks libduckdb-sys without default features, using "bundled-cmake capi-v1" to enable the current v1 bindings.

2.  Removed the added static-HTTPFS test from nightly.rs, leaving that file unchanged relative to upstream. It now lives immediately after test_extension_icu as test_extension_httpfs (crates/duckdb/src/extension.rs:54). The CMake jobs (.github/workflows/rust.yaml:82) enable icu httpfs on Linux/macOS and explicitly execute it. It verifies built-in static loading and an empty extension directory with autoload/ autoinstall disabled.

3. I removed the whole version assertion alongside the unrelated version override. I did not replace it with CARGO_PKG_VERSION derivation, because this test now checks only HTTPFS behavior; source-version reporting remains separate.

4. The README (README.md:281) lists pkg-config, libcurl/OpenSSL development libraries, configure-time network access, installatio commands, and unsupported Windows HTTPFS builds.

deem0n:
@mlafeldt I’ve addressed the review comments and pushed the fixes. All eight CI jobs passed in my fork. The upstream CI run is awaiting maintainer approval. Could you please approve it and take another look when you have time? Thanks!

PR #836, opened 2026-08-29
Update Arrow to 59

geoHeil:
Aligns the workspace with the current `arrow` release. `arrow` 59

## What changed

- `Cargo.toml`: workspace `arrow` requirement `58` → `59`
- `Cargo.lock`: the eleven `arrow-*` crates 58.4.0 → 59.2.0

## Testing

Rust 1.98.0 stable, macOS aarch64, `bundled` backend. 

## Motivation

Downstream workspaces that also depend on `datafusion` currently get two Arrow majors in the tree, since `datafusion` has moved to 59. Arrow is a type vocabulary rather than an ordinary leaf dependency, so a split major is more than a duplicate build cost: anything that wants to hand an Arrow value between `duckdb` and another Arrow consumer needs both sides on the same major.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,9 +104,9 @@ checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
 
 [[package]]
 name = "arrow"
-version = "58.4.0"
+version = "59.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


mlafeldt:
Thanks!

faysou:
It would be great to get a release of duckdb-rs soon to benefit from this sync in Arrow versions without using a develop branch of duckdb-rs which can't be done for every projects.

PR #840, opened 2026-09-02
Add UUID feature to duckdb-rs v2

okruitho:
This pr adds the uuid feature flag to the v2 api. This can be used to idiomatically access uuids from duckdb.


Maxxen:
LGTM but in my experience its a lot easier to feature flag on whole modules/files instead of inline like this as soon as more flags are introduced - particularly when running with strict clippy lints as you otherwise easily get a lot of "unused import" warnings/errors if you dont build with all flags on. But we can move it later.

Also the CI didn't seem to run, so maybe that needs fixing.

PR #842, opened 2026-09-03
Put the libduckdb download client behind a default `download-lib` feature

gurinderu:
## Motivation

`libduckdb-sys` lists `ureq` (with `rustls`) as an unconditional
`[build-dependencies]` entry. Build dependencies are compiled for the host on
every build of every consumer, so each `duckdb` user compiles a TLS stack
(`ureq` → `rustls` → `ring`) for the build script, even with `bundled`, where
the download path is never reached.

This has two costs:

- **It can break builds.** `ring` 0.17.14 (the last release in that line)
  fails its aarch64-apple CPU-feature const-asserts on some macOS CI runners:

  ```
  error[E0080]: evaluation panicked: assertion failed: (CAPS_STATIC & MIN_STATIC_FEATURES) == MIN_STATIC_FEATURES
    --> ring-0.17.14/src/cpu/arm/darwin.rs:44:5
  ```

  A consumer that has already removed `ring` from its own runtime graph still
  cannot build, because Cargo features are additive and there is no way to
  strip a dependency's build-dependency from the outside.

- **It is wasted work.** A `bundled` build compiles the engine from the
  vendored tarball and never touches the network, yet still pays for
  compiling an HTTP client and a TLS provider.

## Changes

Minimal and behaviour-preserving for existing users:

- New `download-lib` feature on `libduckdb-sys` owning `ureq` and `zip`
  (both now `optional`). It is in the default feature set, so default builds
  and `DUCKDB_DOWNLOAD_LIB` keep working exactly as before.
- The `DUCKDB_DOWNLOAD_LIB` code path in `build.rs` is gated on the feature.
  Setting the variable without the feature panics with a message pointing at
  `download-lib`.
- The workspace dependency on `libduckdb-sys` is declared with
  `default-features = false`, and the `duckdb` crate forwards the sys defaults
  through its own `default` feature. Consumers can therefore drop the HTTP
  client with:

  ```toml
  duckdb = { version = "...", default-features = false, features = ["bundled"] }
  ```

- One-line README note. No CI changes, `Cargo.lock` unchanged.

## Verification

- `cargo tree -p duckdb --no-default-features --features bundled --edges normal,build --invert ring`
  reports no `ring` package; with default features `ring` is present via
  `ureq`, as before.
- `cargo clippy -p libduckdb-sys --all-targets -- -D warnings` passes with
  default features and with `--no-default-features --features vcpkg,pkg-config`.
- A full `cargo build -p duckdb --no-default-features --features bundled`
  on aarch64-apple is in progress locally; I will post the result in a comment.

## Compatibility

No change for anyone using default features. Only consumers who opt into
`default-features = false` lose the download path, and they get a clear
build-script error if they also set `DUCKDB_DOWNLOAD_LIB`.


gurinderu:
Follow-up on Verification: `cargo build -p duckdb --no-default-features --features bundled` (aarch64-apple-darwin, Rust 1.98) completed successfully on 76b8eba.

okruitho:
hi @gurinderu, this is definitely an improvement and something we should merge; I've added some changes to your branch, take a look here: https://github.com/gurinderu/duckdb-rs/pull/1 

Thanks!

PR #850, opened 2026-09-08
Expose query progress on Connection

### LINKED ISSUE #207: Get progress of a long running query
Is there a way to poll progress of a long running query?

This is the equivalent c API I found

http://duckdb.org/docs/api/c/api#duckdb_query_progress

sankarreddy-atlan:
DuckDB reports how far the query running on a connection has progressed, and `duckdb_query_progress` is already bound in `libduckdb-sys`, but the safe wrapper never surfaced it. `ClientContext::query_progress` is three atomics read without taking the context lock, so the estimate can be polled from a thread other than the one executing the query. That is the case worth covering: a `Connection` is not `Sync`, so the executing thread owns it for the duration.

- Add `QueryProgress`, `Connection::query_progress`, and a `Send + Sync` `ProgressHandle` that mirrors `InterruptHandle` for the cross-thread case.
- Return `Option<f64>` from `QueryProgress::percentage`, since DuckDB reports -1 for plans it cannot estimate and a caller reading that as 0% would mistake one for a stalled query.
- Add `Connection::set_query_progress_tracking`, off by default. `enable_progress_bar` is a client setting rather than a database option, so `duckdb_set_config` rejects it and it cannot go through `Config`; enabling also clears `enable_progress_bar_print` so a query does not draw a bar on stdout.
- Share one `ConnectionCell` between `InterruptHandle` and `ProgressHandle`, and move `duckdb_disconnect` inside its lock. `close` previously freed the connection before nulling the pointer interrupts read through, so an interrupt racing a close could touch freed memory; a second concurrent reader would have widened that window.

Progress quality varies by source, which the tests account for: a table scan reports an estimate that advances, while a `range()` cross product reports a row total it never makes progress against. Both tests drive a scan through `stream_arrow`, so consuming the stream advances execution and neither races a background query.

`InterruptHandle`'s public API and auto-trait impls are unchanged, with static assertions covering the latter. No semver-breaking changes.

Closes #207, which asked for this to be paired with `duckdb_interrupt` from #493 rather than to wait on the pending API.

### DIFF
--- a/crates/duckdb/src/inner_connection.rs
+++ b/crates/duckdb/src/inner_connection.rs
@@ -59,6 +59,7 @@ pub struct InnerConnection {
     // it is not used for concurrent access to the database itself.
     database: Arc<Mutex<DatabaseHandle>>,
     pub con: ffi::duckdb_connection,
+    pub(crate) cell: Arc<ConnectionCell>,
     interrupt: Arc<InterruptHandle>,
 }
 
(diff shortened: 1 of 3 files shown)


sankarreddy-atlan:
@mlafeldt could you please review this

PR #851, opened 2026-09-08
Better building dynamic

okruitho:
This PR adds an extra package to the repo: duckdb-build. This is an build package to support dynamic linking and libary discovery.

Also updated the README to better reflect the current state of the V2 API.


Maxxen:
So what is the benefit of moving this into a separate package vs just having it in the build.rs of libduckdb-sys?

okruitho:
Even if it is in the same package you will need to have to declare it as a build-dependency and an normal dependency. It's possible and can be done if you want!

For me it seemed like a good split between the two, just like with tauri for example: https://crates.io/crates/tauri-build

okruitho:
This is superceded by #855. closing for now.

PR #855, opened 2026-09-11
Merge the C-API-V2 branch into main, add `duckdb-neo` crate

Maxxen:
This branch merges the `v2.0-cyanoptera` line that @okruitho has been working on back into main. The `libduckdb-sys` crate and build system now serves two wrapper crates, the existing duckdb (V1 C API) and a new `duckdb-rs-neo` which targets the new V2-C-API, against a single (optionally) vendored DuckDB 2.0 alpha build.

**libduckdb-sys**:
 - Bindings for both headers are built from the same source tree. A new `capi-v1` flag (default) generates `duckdb.h` bindings into `libduckdb_sys::v1`, re-exported at the crate root so existing code keeps compiling, while a `capi-v2` flag generates `duckdb_v2.h` bindings into `libduckdb_sys::v2` with rustified enums and an uppercase-aware allowlist, plus the v2 string helpers (orphan rule keeps them here).
  - The cc backend compiles as C++17 as DuckDB 2.0 now requires it, and defines DEBUG or NDEBUG to match DuckDB's CMake build types. update_sources.py excludes third_party/jemalloc, which 2.0 moved into the core list but which is Linux-only - we can consider adding it back in in a separate PR.  
 - Linked mode now downloads a prebuilt DuckDB binary by default. `.duckdb-release` in pins `DUCKDB_RELEASE_COMMIT`, `DUCKDB_RELEASE_URL`, and `DUCKDB_RELEASE_VERSION`. Without DUCKDB_LIB_DIR, the build script fetches `duckdb-shared-libs-<platform>.tar.gz` from that URL and stages the library beside the test binaries. Set `DUCKDB_DOWNLOAD_LIB=0` to opt out. The pin file is updated by the nightly trigger from duckdb-core (and is currently set to an alpha-build of duckdb v2.0). With no pin file it falls back to the GitHub release derived from the crate version. The zip build dependency is gone.
 
### duckdb-rs-neo
  - This is just a  copy of the v2.0-cyanoptera branch wrapper (lib duckdb_rs_neo), depending on the sys crate with default features off and `capi-v2` on. All 57 tests pass against the pinned build. The separate branch can probably be removed after this gets merged as all v2-C-API development should shift to main.
  
### DuckDB bump and v1 fallout
- Vendored DuckDB moves from v1.5.5 to preview commit a00803f768 (v2.0.0-alpha39998); workspace version becomes 1.20000.0
- v1 wrapper test updates for real 2.0 behavior: X'..' literals parse as blobs, test_all_types() gained timestamp_tz_ns and empty_struct (a tuple type is excluded, not modeled), profiling metrics moved into grouped child nodes, and prepared statements re-bind more often.
- Six vscalar error-handling tests are #[ignore]d: DuckDB 2.0 requires scalar functions to be marked fallible, and the v1 C API cannot do that. I'll do an upstream fix separately.

### Tooling
  - `./upgrade.sh --sync` now vendors the pinned commit from `.duckdb-release` and regenerates tarball and bindings, with an optional version argument for release bumps. `--sha` remains for experiments and warns on pin mismatch.
  
### CI
 - rust.yaml: three-OS test job in linked mode against the pinned download, bundled coverage kept in the Linux clippy step and the CMake job, sanitizer back to download-on-PR and bundled-on-main, cargo package instead of                                                                                              
    cargo publish --dry-run, neo added to msrv and sanitizer.
 - nightly.yaml and the new release_version_pr.yml are dispatched by DuckDB's release dispatcher with `duckdb-sha` and `duckdb-version`, the bump workflow rewrites the pin and runs --sync in the same PR. `release.yaml` drops --all-features.
 
 ### Follow-up to do once this is merged
 
 - Register our bump workflows under `core_ready` in `duckdb/duckdb-workflow-trigger`'s `endpoints.yml`.
 - The v1.5 notify workflow still POSTs sha-only dispatches to nightly.yaml and will be rejected until removed there. Make it target the duckdb-rs/v1.5-variegata branch instead of main.
 - ~~The `SetFallible()` fix in core DuckDB's v1 C API upstream.~~ This is now done here! https://github.com/duckdb/duckdb/pull/25664


README.md
```diff
@@ -74,6 +74,37 @@ fn main() -> Result<()> {
...
 Execute the program with `cargo run` and watch DuckDB in action!
 
+### duckdb-neo: the v2 C API wrapper (experimental)
+
+> ⚠ DuckDB's v2 C API is in rapid development. Consider `duckdb-neo` unstable. Feedback is welcome!
+
+`duckdb-neo` is a second wrapper crate in this repository that targets DuckDB's new v2 C API (`duckdb_v2.h`) instead of the v1 API used by `duckdb`. Both crates share the same `libduckdb-sys` crate and therefore the same bundled DuckDB, build options, and version number.
+
+```shell
+cargo add duckdb-neo -F bundled
```
okruitho, line 84: we should probably stop recommending bundled by default with the new download option.

okruitho (approved):
Looks great, thanks!

PR #860, opened 2026-09-17
Wait for the query to start before interrupting in test_interrupt

dajiaohuang:
`test_interrupt` slept for a fixed 100 ms and then interrupted once. That only works if the worker thread has already reached the query's executor by then. DuckDB resets the interrupt flag when a query starts executing, so an interrupt that arrives while the statement is still being prepared is discarded, the query runs to completion, and the test fails on the `recv_timeout` error instead of on the interrupt assertion.

Measured on the pinned build (`v2.0.0-alpha39998`):

- interrupt delivered after `prepare()` returns but before `execute()` starts the executor: discarded, query still running after 5 s
- interrupt delivered while the executor is running: query fails with `INTERRUPT` in ~180 µs

So the outcome depends on whether the worker thread got scheduled in time, which a loaded machine does not guarantee.

The test now signals the interrupt until the query reacts, bounded by a 30 s deadline so an uninterruptible query still fails the test.

### Verification

- `cargo test -p duckdb --lib` before: `test::test_interrupt` failed 3/3 parallel runs (285 passed, 1 failed).
- `cargo test -p duckdb --lib` after: 286 passed, 0 failed, 4/4 runs.
- `cargo test -p duckdb --lib test::test_interrupt -- --exact`: passes in 0.08 s.
- `cargo clippy -p duckdb --lib --tests` and `cargo fmt --all -- --check` clean.


crates/duckdb/src/lib.rs
```diff
@@ -1579,10 +1579,27 @@ mod test {
             tx.send(stmt.execute([])).unwrap();
         });
 
-        std::thread::sleep(std::time::Duration::from_millis(100));
-        db_interrupt.interrupt();
+        // DuckDB resets the interrupt flag when a query starts executing, so an
+        // interrupt that arrives while the worker is still preparing is discarded.
+        // Keep signalling until the query reacts; a single fixed-delay interrupt
+        // is missed whenever the worker is descheduled past the sleep.
+        let deadline = std::time::Instant::now() + std::time::Duration::from_secs(30);
```
okruitho: Is 30 seconds not a bit overkill? Can we bring it back to 5?

okruitho (changes requested):
Thanks! Looks great, one comment ^

crates/duckdb/src/lib.rs
dajiaohuang: Addressed in deb1792: reduced the interrupt-test deadline from 30 seconds to 5 seconds, as suggested. The focused test filter passes (2 tests), and git diff --check is clean.

okruitho (approved):
Thanks!

